Issue #23, opened 2023-06-15
The bug in undo  and redo function.


bullub:
I found that the framework core records all user commands in the `_undoStack`, including non editing operations such as selecting cells.

This results in if I modify the table content and frequently switch between selected cells, and then want to undo the modified content, I will need to click the undo button multiple times.

Dushusir:
Sorry, the core schema is being adjusted and cannot be merged due to conflicts

PR #24, opened 2023-06-15
fix(core,base-sheets): fix undo and redo

bullub:
Special Note: 
Due to the setting of `end_of_line: lf` in editorconfig, there are many invalid changes in `packages/base-sheets/src/Controller/Selection/SelectionManager.ts` and `packages/core/src/Sheets/Domain/Range.ts` files point, hereby explain to modify the position of the line number::

`packages/base-sheets/src/Controller/Selection/SelectionManager.ts`：`20 and 316`
`packages/core/src/Sheets/Domain/Range.ts`:  `15~20  and 3142~3143`

#23


Dushusir:
Sorry, the core schema is being adjusted and cannot be merged due to conflicts

PR #52, opened 2023-08-15
Fix(39): run dev sheet error

boychina:
1. 修复pnpm run dev:sheet报错问题；
2. 修复部分eslint问题；


packages/base-docs/src/View/Render/Views/DocsView.ts
```diff
@@ -18,7 +18,7 @@ export class DocsView extends BaseView {
 
     constructor(
         @Inject(CommandManager) private readonly _commandManager: CommandManager,
-        @ICurrentUniverService private readonly _currentUniverService: ICurrentUniverService,
+        @IDCurrentUniverService private readonly _currentUniverService: ICurrentUniverService,
```
wzhudev, line 21: Please do not rename this service

packages/core/src/Sheets/Domain/Range.ts
```diff
@@ -1,8 +1,9 @@
-import { ICurrentUniverService } from 'src/Service/Current.service';
+import { Inject } from '@wendellhu/redi';
+import { IDCurrentUniverService, ICurrentUniverService } from '../../Service/Current.service';
 
 // TODO@wzhudev: here we still have some
 
-import { Inject } from '@wendellhu/redi';
+// eslint-disable-next-line import/no-cycle
```
wzhudev, line 6: Please do not silent this warning. We intent to fix this later.

packages/core/src/Sheets/Domain/Range.ts
```diff
@@ -347,9 +347,12 @@ export class Range {
      */
     getFontColors(): Array<Array<Nullable<string>>> {
         const styles = this._currentUniverService.getCurrentUniverSheetInstance().getWorkBook().getStyles();
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
wzhudev, line 350: Please do not add these.

wzhudev (changes requested):
Thanks for openning this. Though some problems should be fixed before we merging this.

packages/core/src/Sheets/Domain/Range.ts
boychina: fixed

packages/base-docs/src/View/Render/Views/DocsView.ts
boychina: fixed

wzhudev:
Another name such as "style(*): fix lint errors" would be more appropriate

jikkai:
Already solved this issue in PR #128

PR #85, opened 2023-08-31
refactor(ui): right menu item use command service

Dushusir:


packages/base-ui/src/Interfaces/CustomLabel.ts
```diff
@@ -1,6 +1,21 @@
 import { ComponentChildren } from 'preact';
 import { BaseComponent, JSXComponent } from '../BaseComponent';
 import { CustomComponent } from '../Common';
+import { BaseSelectChildrenProps } from '../Components/Select/Select';
+
+export interface CustomLabelProps {
```
wzhudev, line 6: interface 以 `I` 开头

wzhudev (approved):

PR #121, opened 2023-09-09
refactor(ui): refactoring CanvasIcon

jikkai:


wzhudev:
We don't favor export default because it may cause other developers hard to track the code.

wzhudev (approved):

Issue #123, opened 2023-09-11
House keeping


wzhudev:
The last step to Univer Architecture Ready milestone is to clean up legacy code.

- ~~[ ] fix unit tests (we will do it layer when we implement features)~~
  - [x] setup unit tests environment @jikkai 
- [x] fix type problems & lint problems
- [x] remove legacy system
- [x] remove unused files

PR #127, opened 2023-09-11
chore: fix testing tools for react

jikkai:
related #123


jikkai:
> CI Test failed. Is that expected?

Yes.
We will fix the test cases of `base-ui` later in another pr, as well as temporarily remove the test cases of unfinished plugins

PR #160, opened 2023-09-19
refactor(i18n): redesign i18n

### LINKED ISSUE #70: Redesign i18n

jikkai:
close #70 

- Default language is set to `en`.
- Exported locale modules for each plugin.
- Provided configuration for injecting language packs.
- Introduced a method to modify the language: `setLocale(locale): void => {}`.

TODO:
1. Specification language codes
zh => zh-CN
en => en-US

2. Drop `Shared/Locale.ts`

### DIFF
--- a/examples/univer-sheet-ts/src/main.tsx
+++ b/examples/univer-sheet-ts/src/main.tsx
@@ -1,20 +1,32 @@
-import { DocPlugin } from '@univerjs/base-docs';
+import { DocPlugin, zh as DocPluginZh } from '@univerjs/base-docs';
 import { RenderEngine } from '@univerjs/base-render';
-import { SheetPlugin } from '@univerjs/base-sheets';
-import { UIPlugin } from '@univerjs/base-ui';
+import { SheetPlugin, zh as SheetPluginZh } from '@univerjs/base-sheets';
+import { UIPlugin, zh as UIPluginZh } from '@univerjs/base-ui';
 import { DEFAULT_WORKBOOK_DATA_DEMO } from '@univerjs/common-plugin-data';
 import { LocaleType, Univer } from '@univerjs/core';
-import { FindPlugin } from '@univerjs/sheets-plugin-find';
+import { FindPlugin, zh as FindPluginZh } from '@univerjs/sheets-plugin-find';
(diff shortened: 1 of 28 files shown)


examples/univer-sheet-ts/src/main.tsx
```diff
@@ -1,20 +1,32 @@
...
+import { ImportXlsxPlugin, zh as ImportXlsxPluginZh } from '@univerjs/sheets-plugin-import-xlsx';
+import { NumfmtPlugin, zh as NumberfmtPluginZh } from '@univerjs/sheets-plugin-numfmt';
+import { DocUIPlugin, zh as DocUIPluginZh } from '@univerjs/ui-plugin-docs';
+import { SheetUIPlugin, zh as SheetUIPluginZh } from '@univerjs/ui-plugin-sheets';
 
 // univer
 const univer = new Univer({
     locale: LocaleType.EN,
+    locales: {
+        zh: {
```
wzhudev, line 19: IHMO it is better to extract these into another file? Would be a practice for our users.

packages/base-docs/src/DocPlugin.ts
```diff
@@ -50,7 +50,6 @@ export class DocPlugin extends Plugin<DocPluginObserve> {
     initialize(): void {
         this._localeService.getLocale().load({
```
wzhudev, line 51: Is it still necessary to load i18n files in plugins? Since we have already load i18n files in the top-level Univer.

packages/base-ui/src/views/app.tsx
```diff
@@ -96,6 +96,12 @@ export function App(props: IUniverAppProps) {
 
     const [locale, setLocale] = useState<LocaleType>(localeService.getLocale().getCurrentLocale());
 
+    useEffect(() => {
+        localeService.getLocale().locale$.subscribe((locale) => {
```
wzhudev, line 100: Be careful when you subscribe to an observable. Make sure you will unsubscribe especially ui is upper level to core.

wzhudev (changes requested):
Overall the solution looks good but several issues need to be addressed.

wzhudev (approved):
LGTM

PR #181, opened 2023-09-26
chore(core): using barrel exports

jikkai:
- renamed `services/theme/Theme.ts` to `services/theme/theme.ts`
- moved `services/locale.service.ts` to `services/locale/locale.service.ts`
- using barrel exports for `index.ts` to avoid some storybook issues


wzhudev:
Why switching back to barrel exports? It is hard to control what should be exported.

jikkai:
> Why switching back to barrel exports? It is hard to control what should be exported. Any other solutions to fix storybook issue?

I couldn't find a good way to resolve this issue. 

Whenever I export one of `ICommand`, `ICommandInfo` or `IRowColIter`, the following error occurs:
image

I found a related question on [stackoverflow](https://stackoverflow.com/questions/76424612/storybook-the-requested-module-xxx-does-not-provide-an-export-named-xxx-when-tr), but it seems it hasn't been resolved :(

PR #206, opened 2023-09-28
fix(base-sheets): fix cellstyle bug

tonytonychopper123:
fix(base-sheets): fix cellstyle bug


Dushusir:
Related BUG will be fixed at https://github.com/dream-num/univer/pull/251

PR #255, opened 2023-10-08
feat(base-ui): add `Message` component

jikkai:


packages/base-ui/src/Components/Message/test/index.test.tsx
```diff
@@ -0,0 +1,25 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { describe, expect, test } from 'vitest';
+
+import { Button } from '../../Button';
+import { message } from '..';
```
wzhudev, line 5: 尽量不用 barrel import?

packages/base-ui/src/services/message/desktop-message.service.ts
```diff
@@ -1,10 +1,17 @@
...
 export class DesktopMessageService implements IMessageService {
-    show(): IDisposable {
+    show(options: IShowOptions): IDisposable {
+        const { type, content, delay } = options;
+
+        message[type]({
+            content,
+            delay,
+        });
         return toDisposable(() => {});
```
wzhudev, line 15: Should allow caller to dispose the message.

packages/base-ui/src/services/message/message.service.ts
```diff
@@ -2,6 +2,12 @@ import { createIdentifier, IDisposable } from '@wendellhu/redi';
 
 export const IMessageService = createIdentifier<IMessageService>('univer.message-service');
 
+export interface IShowOptions {
+    type: 'success' | 'warning' | 'error';
```
wzhudev, line 6: Define an enum for it.

wzhudev (approved):

PR #260, opened 2023-10-10
fix(sheet): scroll service no effect

DR-Univer:


packages/base-render/src/Viewport.ts
```diff
@@ -36,6 +36,7 @@ export interface IScrollObserverParam {
     actualScrollY?: number;
     limitX?: number;
     limitY?: number;
+    isTrigger?: boolean;
```
wzhudev, line 39: isTrigger 是啥意思？

PR #280, opened 2023-10-13
test(base-sheets): add undo redo test, set font family, use styles cacheMap

Dushusir:


packages/core/src/sheets/styles.ts
```diff
@@ -42,6 +51,10 @@ export class Styles {
     add(data: IStyleData): string {
         const id = Tools.generateRandomId(6);
         this._styles[id] = data;
+        // update cache
+        const styleObject = JSON.stringify(data);
```
wzhudev, line 55: Should use LRU cache to avoid memory overflow.

wzhudev (approved):

packages/core/src/sheets/styles.ts
Dushusir: Will be fixed in other branches

PR #362, opened 2023-11-01
refactor: remove `CustomLabel` & refactor `NeoCustomLabel`

jikkai:


packages/design/src/components/input-number/InputNumber.tsx
```diff
@@ -58,6 +58,12 @@ export interface IInputNumberProps {
      * @param value
      */
     onChange?: (value: number | null) => void;
+
+    /**
+     * Callback when user press enter
+     * @param e
+     */
+    onPressEnter?: (e: React.KeyboardEvent<HTMLInputElement>) => void;
```
wzhudev, line 66: What's the difference between `onChange` and `onPressEnter`?

jikkai: This is a redundant design that might be used in the future

PR #368, opened 2023-11-02
refactor: remove unused image and formula plugins

Dushusir:


jikkai (approved):

wzhudev:
This brokes `pnpm install`. Because some example monorepos still rely on them. Package.json files should be modified as well.

PR #396, opened 2023-11-06
fix: use ResizeObserver API  for container size changes

jikkai:
Replace the resize event with the **ResizeObserver API**, as the container's size may depend on factors beyond just the window size, such as changes in other DOM elements like opening & closing a sidebar.


Jocs:
Great!

packages/base-render/src/Engine.ts
```diff
@@ -122,11 +121,21 @@ export class Engine extends ThinEngine<Scene> {
         this._container.appendChild(this._canvasEle);
 
         this.resize();
-        this.disposeWithMe(fromWindowEvent('resize', () => this.resize()));
+        const resizeObserver = new ResizeObserver(() => {
+            window.requestIdleCallback(() => {
+                this.resize();
```
wzhudev, line 126: 这里是否需要做一个 debounce 或者 throttle 操作？

packages/base-render/src/Engine.ts
```diff
@@ -122,11 +121,28 @@ export class Engine extends ThinEngine<Scene> {
         this._container.appendChild(this._canvasEle);
 
         this.resize();
-        this.disposeWithMe(fromWindowEvent('resize', () => this.resize()));
+
+        let timer: number | undefined;
+        const resizeObserver = new ResizeObserver(() => {
+            if (!timer) {
+                timer = window.setTimeout(() => {
+                    window.requestIdleCallback(() => {
```
wzhudev, line 129: requestIdleCallback 也会返回一个 number 用于取消 timer，因此无需用 setTimeout 来做调度

jikkai: 👌

wzhudev (approved):

PR #437, opened 2023-11-11
feat(sheet): add intercept type

Gggpound:


packages/core/src/services/sheet-interceptor/utils/index.ts
```diff
@@ -0,0 +1,60 @@
+import { Nullable } from '../../../common/type-utils';
```
wzhudev, line 1: No index.ts

wzhudev (changes requested):

Jocs (approved):

wzhudev (approved):

PR #442, opened 2023-11-12
feat(sheet): statusbar

yuhongz:


packages/ui-plugin-sheets/src/views/status-bar/CopyableStatisticItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+    [key: string]: string;
+};
+
+export const CopyableStatisticItem: React.FC<IStatisticItem> = (item: IStatisticItem) => {
+    const localeService = useDependency(LocaleService);
+    const messageService = useDependency(IMessageService);
+    const formateValue = formatNumber(item.value);
+    const copyToClipboard = async () => {
+        try {
+            await navigator.clipboard.writeText(item.value.toString());
```
wzhudev, line 33: Use clipboard interface here.

packages/ui-plugin-sheets/src/views/status-bar/StatusBar.tsx
```diff
@@ -0,0 +1,142 @@
...
+                        }
+                        return stat;
+                    });
+                    setStatistics(newStatistics);
+                }
+                setIsSingle(newSingleState);
+            }
+        };
+
+        window.addEventListener('resize', handleResize);
```
wzhudev, line 80: @jikkai The UI plugin should expose a service for components to subscribe to layout changes (not only window resize event).

packages/ui-plugin-sheets/src/services/status-bar.service.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export class StatusBarService implements IStatusBarService, IDisposable {
+    private readonly functions = [
+        FUNCTION_NAMES.SUM,
+        FUNCTION_NAMES.MAX,
+        FUNCTION_NAMES.MIN,
+        FUNCTION_NAMES.AVERAGE,
+        FUNCTION_NAMES.COUNT,
+    ];
+    private readonly _state$ = new BehaviorSubject<Nullable<IStatusBarServiceStatus>>(null);
```
wzhudev, line 27: IMHO the status should be stored in the local storage and get loaded from the local storage when the web page refreshes.

packages/ui-plugin-sheets/src/services/status-bar.service.ts
```diff
@@ -0,0 +1,47 @@
...
+    getFunctions(): Readonly<FUNCTION_NAMES[]>;
+}
+
+export type IStatusBarServiceStatus = Array<{
+    func: FUNCTION_NAMES;
+    value: number;
+}>;
+
+export class StatusBarService implements IStatusBarService, IDisposable {
+    private readonly functions = [
```
wzhudev, line 20: Privde fields' names should start with an underbar.

packages/ui-plugin-sheets/src/controllers/status-bar.controller.ts
```diff
@@ -0,0 +1,87 @@
...
+    private _registerSelectionListener(): void {
+        this._selectionManagerService.selectionInfo$.subscribe((selectionRangeWithStyle) => {
+            if (selectionRangeWithStyle) {
+                this._calculateSelection(selectionRangeWithStyle);
+            }
+        });
+    }
+
+    private _calculateSelection(selections: ISelectionWithStyle[]) {
+        const unitId = this._univerInstanceService.getCurrentUniverSheetInstance().getUnitId();
```
wzhudev, line 36: There are redundant calls to `this._univerInstanceService.getCurrentUniverSheetInstance()`. You only have to call it once.

packages/base-formula-engine/src/value-object/primitive-object.ts
```diff
@@ -204,7 +204,12 @@ export class NumberValueObject extends BaseValueObject {
         if (valueObject.isArray()) {
             return valueObject.plus(this);
         }
-        return this.plusBy(valueObject.getValue());
+        const object = this.plusBy(valueObject.getValue());
```
wzhudev, line 207: @DR-Univer Please make sure this is appropriate.

wzhudev (changes requested):
And some interaction problems need to be discussed.

packages/base-formula-engine/src/value-object/primitive-object.ts
yuhongz: This part of code was provided by @DR-Univer 🤣

packages/ui-plugin-sheets/src/services/status-bar.service.ts
yuhongz: The configuration will not be cached locally. Actually, status bar will display three or one statistical item by default based on the current size. Reference -> StatusBar.tsx 😄

wzhudev (approved):

PR #454, opened 2023-11-15
fix(sheet): sheet bar scroll button's state

Dushusir:
- [x] sheet bar left and right buttons' state
- [x] slide tab bar init event listener
- [x] rename empty check alert
- [x] fix rename trigger input edit

### DIFF
--- a/packages/ui-plugin-sheets/src/commands/commands/rename.command.ts
+++ b/packages/ui-plugin-sheets/src/commands/commands/rename.command.ts
@@ -13,7 +13,6 @@ export const RenameSheetOperation: ICommand = {
     handler: async (accessor: IAccessor, params?: IRenameCommandParams) => {
         const sheetBarService = accessor.get(ISheetBarService);
         if (params) {
-            console.info('rename ===params', params);
             sheetBarService.setRenameId(params.worksheetId);
         }
         return true;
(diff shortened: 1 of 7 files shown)


packages/ui-plugin-sheets/src/commands/commands/rename.command.ts
```diff
@@ -13,6 +13,7 @@ export const RenameSheetOperation: ICommand = {
     handler: async (accessor: IAccessor, params?: IRenameCommandParams) => {
         const sheetBarService = accessor.get(ISheetBarService);
         if (params) {
+            console.info('rename ===params', params);
```
wzhudev, line 16: Remove console.

packages/ui-plugin-sheets/src/views/sheet-bar/sheet-bar-tabs/SheetBarTabs.tsx
```diff
@@ -79,26 +85,33 @@ export function SheetBarTabs() {
             onScroll: (state: IScrollState) => {
                 sheetbarService.setScroll(state);
             },
+            onEmptyAlert: (message: string) => {
+                confirmService.open({
+                    id: 'slideTabBarAlert',
```
wzhudev, line 90: Extract a const string here.

wzhudev (changes requested):

packages/ui-plugin-sheets/src/commands/commands/rename.command.ts
Dushusir: > Remove console.
Embarrassed

Dushusir: > Remove console.

Embarrassed

wzhudev (approved):

Issue #444, opened 2023-11-13
[Bug] sheet delete selection error


weird94:
### Initial checklist

- [X] Is this really a problem?
- [X] I have searched the Github Issues for similar issues, but did not find anything.

### Affected packages and versions

base-sheets

### Steps to reproduce

1. select more than one cell
2. click delete key

### Expected behavior

the content of the selection need to be deleted

### Actual behavior

only the content on the first cell of this selection was deleted

### Runtime

Chrome

### OS

macOS

PR #467, opened 2023-11-16
fix(sheet): selection delete

weird94:
#444 
fix backspace shortcut in sheet
from `sheet.operation.set-cell-edit-visible` to `sheet.command.clear-selection-content`


weird94:
修改了一下sheet选区删除的command，改成了 `sheet.command.clear-selection-content` 行为和google docs一致，辛苦看下

PR #513, opened 2023-11-22
fix(doc): no need to refresh selection

Jocs:
- [x] no need to refresh selection


wzhudev:
The problems still exists in the preview https://univer-f1wom7ztp-jikkai.vercel.app/sheet/

wzhudev (approved):

PR #522, opened 2023-11-22
feat(doc): support use arrow up and down to move cursor to the beginning and end of doc

Jocs:
- [x] support use arrow up and down to move cursor to the beginning and end of doc

### DIFF
--- a/packages/base-docs/src/controllers/move-cursor.controller.ts
+++ b/packages/base-docs/src/controllers/move-cursor.controller.ts
@@ -69,11 +69,11 @@ export class MoveCursorController extends Disposable {
 
                 switch (command.id) {
                     case MoveCursorOperation.id: {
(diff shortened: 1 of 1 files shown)


packages/base-docs/src/controllers/move-cursor.controller.ts
```diff
@@ -149,6 +148,22 @@ export class MoveCursorController extends Disposable {
 
             const newPos = this._getTopOrBottomPosition(skeleton, focusSpan, direction === Direction.DOWN);
             if (newPos == null) {
+                // move selection
+                const newFocusOffset = direction === Direction.UP ? 0 : dataStreamLength - 2;
```
wzhudev, line 152: 为什么这里 -2?

DR-Univer (approved):

PR #553, opened 2023-11-25
feat: dbclick to fill range

yuhongz:


packages/ui-plugin-sheets/src/controllers/auto-fill.controller.ts
```diff
@@ -49,6 +49,34 @@ export class AutoFillController extends Disposable {
         this._init();
     }
 
+    private _detectFillRange(sourceRange: IRange) {
```
wzhudev, line 52: Unit test required.

wzhudev (changes requested):

wzhudev (approved):

PR #620, opened 2023-12-04
fix(sheet): need empty history stack when cell editor lose focus

Jocs:
- [x] need empty history stack when cell editor lose focus

### DIFF
--- a/packages/sheets-ui/src/controllers/editor/editing.controller.ts
+++ b/packages/sheets-ui/src/controllers/editor/editing.controller.ts
@@ -1,24 +1,21 @@
-import type { Nullable } from '@univerjs/core';
 import {
-    Disposable,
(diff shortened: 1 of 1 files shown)


packages/sheets-ui/src/controllers/editor/editing.controller.ts
```diff
@@ -29,6 +38,22 @@ export class EditingController extends Disposable {
 
     private _initialize() {
         this._initialNormalInput();
+        this._listenEditorBlur();
+    }
+
+    private _listenEditorBlur() {
+        this._currentUniverService.currentDoc$.subscribe((docDataModel) => {
```
wzhudev, line 45: This listener will not be disposed by the subscriber.

wzhudev (approved):

PR #662, opened 2023-12-06
feat(core): add `logLevel` flag to toggle the switch of log service

jikkai:
~~Doesn't seem like the most suitable configuration name, any better suggestions?~~

### DIFF
--- a/packages/core/src/services/log/log.service.ts
+++ b/packages/core/src/services/log/log.service.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-magic-numbers */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
 import { createIdentifier } from '@wendellhu/redi';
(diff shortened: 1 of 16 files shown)


packages/core/src/services/log/log.service.ts
```diff
@@ -1,25 +1,32 @@
...
 
 export const ILogService = createIdentifier<ILogService>('univer.log');
 
 export class DesktopLogService extends Disposable implements ILogService {
-    private _logEnabled = true;
+    private _logLevel: LogLevel = LogLevel.SILENT;
 
     log(...args: any[]): void {
-        if (!this._logEnabled || !args.length) {
+        if (this._logLevel < LogLevel.VERBOSE || !args.length) {
```
wzhudev, line 29: `<=` ?

wzhudev (approved):

PR #696, opened 2023-12-09
refactor(doc): text selection

Jocs:
- [x] add subUnitId\segmentId to text selection manager
- [x] Pls use textSelectionManager to get selections and activeTextRange

### DIFF
--- a/packages/sheets-formula/src/controllers/formula-input.controller.ts
+++ b/packages/sheets-formula/src/controllers/formula-input.controller.ts
@@ -25,6 +25,7 @@ import {
     serializeRange,
     toDisposable,
 } from '@univerjs/core';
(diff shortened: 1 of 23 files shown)


packages/sheets-formula/src/controllers/formula-input.controller.ts
```diff
@@ -25,6 +25,7 @@ import {
     serializeRange,
     toDisposable,
 } from '@univerjs/core';
+import { TextSelectionManagerService } from '@univerjs/docs/services/text-selection-manager.service.js';
```
wzhudev, line 28: 注意导出路径

Jocs: ok

PR #740, opened 2023-12-13
feat(sheet): reduce style size

Gggpound:


Gggpound:
thie work mv to univer-pro
https://github.com/dream-num/univer-pro/pull/110/files

PR #743, opened 2023-12-13
feat(design): add selected icon for tree component

Dushusir:


packages/design/src/components/tree/Tree.tsx
```diff
@@ -138,13 +138,20 @@ export function Tree(props: ITreeProps) {
...
                     })}
                 >
-                    <span
-                        className={clsx(styles.treeIcon, {
-                            [styles.treeIconExpand]: _expand,
-                        })}
-                    >
-                        {children && <DropdownSingle />}
-                    </span>
+                    {children && children.length > 0 && (
```
jikkai, line 141: `children?.length &&`

jikkai (approved):

packages/design/src/components/tree/Tree.tsx
Dushusir: it will get 0 on web page

PR #750, opened 2023-12-14
fix(doc): support copy plain text and without html

Jocs:

close #


packages/docs/src/services/clipboard/clipboard.service.ts
```diff
@@ -60,32 +60,47 @@ export class DocClipboardService extends Disposable implements IDocClipboardServ
...
+                if (copyCache) {
+                    return copyCache;
+                }
             }
-        }
 
-        return this._htmlToUDM.convert(html);
+            return this._htmlToUDM.convert(html);
+        } catch (e) {
+            console.error(e);
```
wzhudev, line 101: Use logService instead of console.log.

wzhudev (approved):
LGTM

packages/docs/src/services/clipboard/clipboard.service.ts
Jocs: It just used for debug, I'll remove it.

PR #768, opened 2023-12-15
refactor(core): support lazy load plugins

wzhudev:

close #


packages/docs/src/controllers/page-render.controller.ts
```diff
@@ -58,6 +58,7 @@ export class PageRenderController extends Disposable {
 
             const docsComponent = mainComponent as Documents;
 
+            console.log('debug', docsComponent.getSkeleton());
```
jikkai, line 61: 😈

wzhudev: Good catch!

PR #989, opened 2024-01-03
fix(sheet): too much time complexity to fix

Gggpound:


Gggpound:
In' Sheet5' , select the first row to merge cells, then drag to the bottom and drag.
Time consuming before optimization

PR #1053, opened 2024-01-09
fix(sheet): resource should be disposed when univer dispose

Gggpound:
- [ ] I am sure that the code is update-to-date with the `dev` branch.

close #


wzhudev (approved):
LGTM

PR #1297, opened 2024-01-30
refactor: optimization of TextX codes

Jocs:

close #


packages/core/src/docs/data-model/text-x/__tests__/utils.spec.ts
```diff
@@ -339,4 +341,39 @@ describe('test text-x utils', () => {
             }],
         });
     });
+
+    it('test isUselessRetainAction()', () => {
```
wzhudev, line 345: Should describe expected results in `it`.

wzhudev (approved):

packages/core/src/docs/data-model/text-x/__tests__/utils.spec.ts
Jocs: I think expect expresion already explain the expect result, it also has different expected in this case.

PR #1328, opened 2024-02-03
feat(core): set appVersion to package version

jikkai:


packages/core/src/types/const/const.ts
```diff
@@ -72,7 +73,7 @@ export const DEFAULT_WORKBOOK: IWorkbookData = {
     id: '',
     sheetOrder: [],
     name: '',
-    appVersion: '',
+    appVersion: version,
```
wzhudev, line 76: 这个字段之前的语义是应用的版本号的话，snapshot schema 也需要一个版本号

wzhudev (approved):

PR #1468, opened 2024-03-04
fix(core): 注释拼写问题

lvanboy:


jikkai:
Please change the title of the PR to conform to the [conventional commits specification](https://www.conventionalcommits.org/en/v1.0.0/).

jikkai:
Thank you for your PR. Since comments should be in English, we have changed the comment here in #1469

PR #1499, opened 2024-03-06
fix(editor): cursor and menu and single choice

DR-Univer:
- [x] cursor position fix
- [x] menu destroy fix
- [x] single range choice
- [x] reduce frequence for value change 

close #


wzhudev (approved):
LGTM!

PR #1564, opened 2024-03-13
feat(sheets): support ref-range-formula-service & custom-cell-render-extension

weird94:
- [ ] I am sure that the code is update-to-date with the `dev` branch.

close #


packages/engine-render/src/components/sheets/extensions/custom.ts
```diff
@@ -0,0 +1,74 @@
...
+            const cellData = worksheet.getCell(row, col);
+            if (cellData && cellData?.customRender) {
+                const primaryWithCoord = this.getCellIndex(row, col, rowHeightAccumulation, columnWidthAccumulation, dataMergeCache);
+
+                if (primaryWithCoord.isMerged && !primaryWithCoord.isMergedMainCell) {
+                    return;
+                }
+
+                // current cell is hidden
+                if (!worksheet.getColVisible(col) || !worksheet.getRowVisible(row)) {
```
DR-Univer, line 50: 贴图渲染缓存要在这里考虑下
image

packages/sheets-formula/src/services/formula-ref-range.service.ts
```diff
@@ -0,0 +1,112 @@
...
+export class FormulaRefRangeService extends Disposable {
+    constructor(
+        @Inject(RefRangeService) private readonly _refRangeService: RefRangeService,
+        @Inject(LexerTreeBuilder) private readonly _lexerTreeBuilder: LexerTreeBuilder,
+        @IUniverInstanceService private readonly _univerInstanceService: IUniverInstanceService
+    ) {
+        super();
+    }
+
+    registerFormula(formulaId: string, formula: string, callback: FormulaChangeCallback): IDisposable {
```
DR-Univer, line 46: update-formula.controller.ts 里的逻辑有做抽象么？这里没有看到，最好是一套逻辑来维护

packages/core/src/types/interfaces/i-cell-custom-render.ts
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { Nullable } from '../../shared';
+import type { IStyleData } from './i-style-data';
+import type { ICellDataForSheetInterceptor } from './i-cell-data';
+import type { ISelectionCellWithCoord } from './i-selection-data';
+
+export interface ICellRenderContext {
```
wzhudev, line 22: These interface may be not the most appropriate to be placed in package `core`.

packages/core/src/types/interfaces/i-cell-data.ts
```diff
@@ -48,6 +49,7 @@ export interface ICellData {
 export interface ICellDataForSheetInterceptor extends ICellData {
     interceptorStyle?: Nullable<IStyleData>;
     isInArrayFormulaRange?: Nullable<boolean>;
+    customRender?: Nullable<ICellCustomRender[]>;
```
wzhudev, line 52: We should provide a similar interceptor on the rendering package instead of making into core.

wzhudev (changes requested):
JSDoc is not provided. IMHO there should be examples to demonstrate how to use this API.

PR #1630, opened 2024-03-18
ci: add complexity check

wzhudev:
https://github.com/eslint/eslint/issues/4808

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -23,6 +23,13 @@ export default antfu({
     rules: baseRules,
 }, {
     files: ['**/*.ts', '**/*.tsx'],
+    ignores: ['**/*.d.ts', '**/vite.config.ts', 'playwright.config.ts', '**/*.spec.ts', '**/*.spec.tsx'], // do not check test files
+    rules: {
+        complexity: ['warn', { max: 20 }],
+        'max-lines-per-function': ['warn', 80],
+    },
+}, {
+    files: ['**/*.ts', '**/*.tsx'],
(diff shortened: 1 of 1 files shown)


eslint.config.js
```diff
@@ -22,7 +22,41 @@ export default antfu({
     rules: baseRules,
 }, {
     files: ['**/*.ts', '**/*.tsx'],
-    ignores: ['**/*.d.ts', '**/vite.config.ts', 'playwright.config.ts'],
+    ignores: ['**/*.d.ts', '**/vite.config.ts', 'playwright.config.ts', '**/*.spec.ts', '**/*.spec.tsx`'],
+    plugins: {
+        header,
+    },
+    rules: {
+        'header/header': [
```
jikkai, line 30: It is not necessary to define the header rule twice in separate configurations, as it can be defined once in the main configuration.

eslint.config.js
```diff
@@ -22,7 +22,41 @@ export default antfu({
...
+                ' * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.',
+                ' * See the License for the specific language governing permissions and',
+                ' * limitations under the License.',
+                ' ',
+            ],
+            2,
+        ],
+
+        // code readability
+        complexity: ['error', { max: 20 }],
```
jikkai, line 54: The new changes can have a significant impact on the codebase. 🤯

wzhudev: Yes. Let's make it a warning for now.

eslint.config.js
wzhudev: OK. Learned something new!

jikkai (approved):

PR #1635, opened 2024-03-19
feat(sheets): add watch API to ref-range-service

wzhudev:
close #xxx, #yyy, #zzzz


packages/sheets/src/services/ref-range/__tests__/watchRange.spec.ts
```diff
@@ -0,0 +1,435 @@
+/**
```
wzhudev, line 1: Be careful with the file's name. It should be in kebab-case.

packages/sheets/src/services/ref-range/ref-range.service.ts
```diff
@@ -64,6 +103,38 @@ export class RefRangeService extends Disposable {
         });
     }
 
+    watchRange(unitId: string, subUnitId: string, range: IRange, callback: WatchRangeCallback): IDisposable {
```
wzhudev, line 106: This API seems not tested.

wzhudev (commented):
Some issue need to be addressed. @ybzky

wzhudev (commented):
Some issue need to be addressed. @ybzky

PR #1665, opened 2024-03-22
fix(sheets-ui): merge setRangeValuesMutation at clipboardservice

yuhongz:
close #xxx, #yyy, #zzzz

### DIFF
--- a/packages/sheets-ui/src/services/clipboard/__tests__/utils.spec.ts
+++ b/packages/sheets-ui/src/services/clipboard/__tests__/utils.spec.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import type { IRange } from '@univerjs/core';
+import type { IMutationInfo, IRange } from '@univerjs/core';
 import { SetRangeValuesMutation } from '@univerjs/sheets';
 
 import { describe, expect, it } from 'vitest';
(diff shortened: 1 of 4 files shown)


packages/sheets-ui/src/services/clipboard/__tests__/utils.spec.ts
```diff
@@ -134,3 +136,60 @@ describe('test getRepeatRange', () => {
         ]);
     });
 });
+
+describe('test "mergeSetRangeValues"', () => {
+    it('empty', () => {
```
wzhudev, line 141: The cases' names should be clear and describe your expections.

packages/sheets-ui/src/services/clipboard/utils.ts
```diff
@@ -99,3 +102,52 @@ export async function clipboardItemIsFromExcel(html: string): Promise<boolean> {
...
+            const matrix = new ObjectMatrix(cellValue);
+            matrix.forValue((row, col, value) => {
+                newMatrix.setValue(row, col, { ...newMatrix.getValue(row, col), ...value });
+            });
+        }
+    });
+    return newMatrix.getMatrix();
+}
+
+export function setRangeValuesMergeable(m1: IMutationInfo<ISetRangeValuesMutationParams>, m2: IMutationInfo<ISetRangeValuesMutationParams>) {
```
wzhudev, line 122: There is not assigning in this function. Should be `getRangesValuesMergeable`?

wzhudev (changes requested):

wzhudev (approved):

PR #1683, opened 2024-03-25
fix(ui): canvas popup event bind error

weird94:
close #xxx, #yyy, #zzzz


packages/sheets-ui/src/services/canvas-pop-manager.service.ts
```diff
@@ -244,11 +245,49 @@ export class CanvasPopManagerService extends Disposable {
             anchorRect$: position$,
         });
 
+        const onDispose = () => {
+            this._globalPopupManagerService.removePopup(id);
+            disposable.dispose();
+            position$.complete();
+        };
+
+        const commandDisposable = this._commandService.onCommandExecuted((commandInfo) => {
```
wzhudev, line 254: It is better to integrate `RefRangeService` here in the future. cc @Gggpound

wzhudev (approved):
LGTM. Some improvements can be made in the future.

PR #1716, opened 2024-03-27
fix(ui): fix toolbar responsiveness on small screens

jikkai:


packages/ui/src/views/components/doc-bars/Toolbar.tsx
```diff
@@ -94,29 +112,31 @@ export function Toolbar(props: IToolbarProps) {
     useEffect(() => {
         function resize() {
             const wrapperWidth = toolbarRef.current?.clientWidth ?? 0;
-            const containerWidth = toolbarContainerRef.current?.clientWidth ?? 0;
+            const gap = 8;
```
Jocs, line 115: Prefer use uppercase. `GAP`

Jocs (approved):

PR #1737, opened 2024-03-29
feat(sheet): defined name

DR-Univer:


packages/engine-formula/src/basics/function.ts
```diff
@@ -116,6 +116,11 @@ export enum FunctionType {
      * User-defined functions
      */
     User,
+
+     /**
+      * Defined name
+      */
+    definedName,
```
Dushusir, line 123: DefinedName

packages/sheets-ui/src/controllers/selection.controller.ts
```diff
@@ -95,6 +96,81 @@ export class SelectionController extends Disposable {
         });
     }
 
+    private _iniDefinedNameListener() {
```
Dushusir, line 99: _initDxxxxxxx

Dushusir (approved):

packages/sheets-ui/src/locale/en-US.ts
```diff
@@ -490,9 +490,35 @@ const locale: typeof zhCN = {
             'set-strike-through': 'Toggle strike through',
         },
     },
-
     'sheet-view': 'Sheet View',
     'sheet-edit': 'Sheet Edit',
+
+    definedName: {
```
jikkai, line 496: Should use kebab-case

jikkai (changes requested):

jikkai (approved):

PR #1742, opened 2024-03-29
fix(sheets): fix shallow copy bugs of the sheet snapshot

karlsbeard:
* fix: change the _snapshot of the workbook to deep clone to avoid the side effects when you new an empty sheet.

### DIFF
--- a/packages/core/src/sheets/workbook.ts
+++ b/packages/core/src/sheets/workbook.ts
@@ -20,7 +20,7 @@ import { ILogService } from '../services/log/log.service';
 import type { Nullable } from '../shared';
 import { Tools } from '../shared';
 import { Disposable } from '../shared/lifecycle';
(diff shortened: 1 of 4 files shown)


packages/core/src/sheets/workbook.ts
```diff
@@ -87,7 +87,7 @@ export class Workbook extends Disposable {
     ) {
         super();
 
-        this._snapshot = Tools.commonExtend(DEFAULT_WORKBOOK, workbookData);
+        this._snapshot = Tools.deepClone(Tools.commonExtend(DEFAULT_WORKBOOK, workbookData));
```
jikkai, line 90: Either perform a deep clone of `DEFAULT_WORKBOOK` or create a `getEmptySnapshot` function, similar to what is done in `packages/core/src/docs/data-model/empty-snapshot.ts`.

karlsbeard: Got it and have solved it.

jikkai (approved):

PR #1762, opened 2024-04-02
feat(facade): add getSheetBySheetId API on FWorkbook

fmwww:

### DIFF
--- a/packages/facade/src/apis/sheets/f-worksheet.ts
+++ b/packages/facade/src/apis/sheets/f-worksheet.ts
@@ -31,6 +31,10 @@ export class FWorksheet {
         @ICommandService private readonly _commandService: ICommandService
     ) {}
 
+    getId(): string {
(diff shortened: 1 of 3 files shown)


Dushusir (approved):

packages/facade/src/apis/sheets/f-worksheet.ts
```diff
@@ -31,6 +31,10 @@ export class FWorksheet {
         @ICommandService private readonly _commandService: ICommandService
     ) {}
 
+    getId(): string {
```
hexf00, line 34: In order to reduce the learning cost for facade API users, it is expected to align with the Google Apps Script API, https://developers.google.com/apps-script/reference/spreadsheet/sheet#getSheetId()

hexf00: resolved on #1770

PR #1770, opened 2024-04-02
feat(facade): sheet api getId rename to getSheetId

hexf00:
In order to reduce the learning cost for facade API users, it is expected to align with the Google Apps Script API, https://developers.google.com/apps-script/reference/spreadsheet/sheet#getSheetId()

#1762


wzhudev (commented):
LGTM

PR #1773, opened 2024-04-02
style(design): add color and text-decoration-line on hover condition

karlsbeard:
* style: add the color and text decoration on the hover condition of the slider


karlsbeard:
done the fix

jikkai (approved):

PR #1775, opened 2024-04-02
feat(facade): add getMaxColumns and getMaxRows API on FWorksheet

fmwww:

### DIFF
--- a/packages/facade/src/apis/sheets/f-worksheet.ts
+++ b/packages/facade/src/apis/sheets/f-worksheet.ts
@@ -93,7 +93,7 @@ export class FWorksheet {
 
     /**
      * Returns the current number of columns in the sheet, regardless of content.
-     * @return The maximum width of the sheet
+     * @return The maximum columns count of the sheet
      */
     getMaxColumns() {
         return this._worksheet.getMaxColumns();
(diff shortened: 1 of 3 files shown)


packages/facade/src/apis/sheets/f-worksheet.ts
```diff
@@ -90,4 +90,20 @@ export class FWorksheet {
             colWidth: width,
         } as ISetWorksheetColWidthMutationParams);
     }
+
+    /**
+     * Returns the current number of columns in the sheet, regardless of content.
+     * @return The maximum width of the sheet
```
jikkai, line 96: Maybe it's not related to the width or the height?

jikkai (changes requested):

packages/facade/src/apis/sheets/f-worksheet.ts
fmwww: 我看GoogleSheet那边是这么定义的 [https://developers.google.com/apps-script/reference/spreadsheet/sheet#getmaxcolumns](url)

jikkai: This might be an mistake, and insisting on using `width` and `height` could lead to confusion.

hexf00 (approved):

jikkai (approved):
LGTM

PR #2067, opened 2024-04-24
fix: fix current render handling in desktop controller

wzhudev:


packages/ui/src/controllers/ui/ui-desktop.controller.tsx
```diff
@@ -76,6 +76,8 @@ export class DesktopUIController extends Disposable implements IDesktopUIControl
                 this._renderManagerService.currentRender$.subscribe((renderId) => {
                     if (renderId) {
                         const render = this._renderManagerService.getRenderById(renderId)!;
+                        if (!render.unitId) return;
```
Jocs, line 79: Under what circumstances does unitId not exist?

wzhudev: Cell editors.

jikkai (approved):

PR #2073, opened 2024-04-25
refactor: make the source of univer-instance-type unique

hexf00:
1. Unify the usage of `UniforInstanceType`
2. Avoid misuse of `@univerjs/protocol`

Unless directly used at the network interface layer, types should be introduced from the `@univerjs/core`

```
pnpm build
```

and Don't use  `UNRECOGNIZED` type , this is a generation error, may be fixed in the future, existing usage will be removed in other PRs


wzhudev (approved):

wzhudev:
Merge this later after all other changes in 0.1.8. @hexf00 Please prepare to rebase on the latest dev.

Jocs (approved):

PR #2100, opened 2024-04-28
fix: global-zone componentKey initial value error

weird94:
close #xxx, #yyy, #zzzz


Dushusir (approved):

packages/ui/src/services/global-zone/desktop-global-zone.service.ts
```diff
@@ -25,15 +25,22 @@ import type { IGlobalZoneService } from './global-zone.service';
 export class DesktopGlobalZoneService implements IGlobalZoneService {
     readonly visible$ = new Subject<boolean>();
     readonly componentKey$ = new Subject<string>();
```
wzhudev, line 27: Make this a `BehaviorSubject` and give it an initial value would be simpler.

weird94: get

PR #2107, opened 2024-04-29
fix(conditional-formatting): viewmodel interception is not required a…

Gggpound:
1. In version 42, for example, 10 custom formula conditional formats were created. 
2. In version 50, snapshot save.
3. In version 51, remove 10 custom formula conditional format 
4. Other actions to version 70. 
5. When the user enters version 70, they will pull the CS of  50-70 and version of snapshot 50.
6. The snapshot of version 50 will have a conditional format, and when applying version 50-70,  deleting the 10 custom formulas.
7. Conditional formatting should not be triggered when applied

### DIFF
--- a/packages/core/src/shared/after-init-apply.ts
+++ b/packages/core/src/shared/after-init-apply.ts
@@ -1,0 +1,35 @@
+/**
+ * Copyright 2023-present DreamNum Inc.
+ *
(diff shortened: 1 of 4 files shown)


DR-Univer (approved):

wzhudev (approved):
Looks clean.

Issue #2169, opened 2024-05-07
[Bug] Select a hidden row and then add a record, the hidden icon is displayed incorrectly


oumomomo:
### Before you submit this issue, have you checked the following

- [X] Is this really a problem?
- [X] I have searched the Github Issues for similar issues, but did not find anything.

### Affected packages and versions

dev

### Reproduction link
问题1、
操作步骤：
1、隐藏部分记录/列
2、选中隐藏行/列+普通行/列
3、新增记录/列

问题2、
1、隐藏部分行/列
2、选中隐藏行/列+普通行列并删除

### Expected behavior
问题1、
隐藏的icon应展示在对应的隐藏位置，点击取消隐藏也不会再展示第二个隐藏icon
问题2、
隐藏行/列被删除后，隐藏icon消失

### Actual behavior
问题1、
新增记录后隐藏的icon位置展示有误，取消隐藏后还会有第二个icon
[2024_5_7 14_51_02 video.webm](https://github.com/dream-num/univer/assets/167171356/79ac2977-bf89-4517-9827-228bda39ea1d)
[2024_5_7 14_51_02 commands.json](https://github.com/dream-num/univer/files/15231505/2024_5_7.14_51_02.commands.json)

问题2、
隐藏行/列被删除后，隐藏icon未消失
[2024_5_7 15_50_25 commands.json](https://github.com/dream-num/univer/files/15232076/2024_5_7.15_50_25.commands.json)
[2024_5_7 15_50_25 video.webm](https://github.com/dream-num/univer/assets/167171356/0e15da95-36b2-41ad-9652-1f7e30a5959d)

PR #2203, opened 2024-05-09
fix(sheet): expand hidden icon render incorrect when rows and cols ch…

siam-ese:
…anged

close [2169](https://github.com/dream-num/univer/issues/2169#event-12744250664)


siam-ese:
@wzhudev, I'm not sure if this list is full, are there any other commands I'm missing?

PR #2208, opened 2024-05-09
fix(sheet): unhide hidden icon render incorrect when headers resize

siam-ese:
close [#2169](https://github.com/dream-num/univer/issues/2169#event-12744264054)


wzhudev (approved):
LGTM

PR #2215, opened 2024-05-10
feat(sheet): allow menu scroll when it over viewport

siam-ese:
To resolve the below problem

WeChatWorkScreenshot_9b0d8287-a901-45dc-9e0a-5e71b988f74c

The height of the dropdown menu on the Toolbar is greater than the height of the viewport. 
It brings an unexpected scroller.


jikkai (approved):
LGTM

PR #2221, opened 2024-05-11
fix(sheet): skip filtered row on setting style

siam-ese:
close [#685](https://github.com/dream-num/univer-pro/issues/685#event-12768432550)

### DIFF
--- a/packages/sheets/src/commands/commands/utils/selection-utils.ts
+++ b/packages/sheets/src/commands/commands/utils/selection-utils.ts
@@ -216,7 +216,7 @@ export function isSingleCellSelection(selection: Nullable<ISelectionWithStyle & 
  * @param sheet bind a sheet
  * @returns iterator
  */
-export function createRangeIterator(sheet: Worksheet) {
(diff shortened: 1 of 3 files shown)


packages/sheets/src/commands/commands/utils/selection-utils.ts
```diff
@@ -210,3 +210,27 @@ export function isSingleCellSelection(selection: Nullable<ISelectionWithStyle &
     const { range, primary } = selection;
     return Rectangle.equals(range, primary);
 }
+/**
+ * Create an iterator to iterate over cells in range.
+ * It will skip the row that has been filtered.
+ * @param sheet bind a sheet
+ * @returns iterator
+ */
+export function createRangeIterator(sheet: Worksheet) {
```
wzhudev, line 219: 把会跳过筛选列的逻辑蕴含在方法名字当中

wzhudev (approved):
The solution looks good to me. The name of the util function should be changed.

PR #2223, opened 2024-05-11
fix(sheet): filterRenderController should mount to rendered stage

siam-ese:
close [#2108](https://github.com/dream-num/univer/issues/2108#event-12775365071)


packages/sheets-ui/src/services/selection/selection-render.service.ts
```diff
@@ -931,7 +931,7 @@ export class SelectionRenderService implements ISelectionRenderService {
         if (scene == null || skeleton == null) {
             return;
         }
-        const { scaleX, scaleY } = scene.getAncestorScale();
+        // const { scaleX, scaleY } = scene.getAncestorScale();
```
siam-ese, line 934: unused code

PR #2245, opened 2024-05-15
fix: the content is not display when enter edit cell

Jocs:
close https://github.com/dream-num/univer-pro/issues/741


wzhudev (approved):
LGTM.

PR #2275, opened 2024-05-18
feat(ui)!: add support for hiding context menu

jikkai:
BREAKING CHANGE: The default values for `header` and `footer` are now set to true.


wzhudev (approved):
lgtm

PR #2357, opened 2024-05-30
feat(facade): add lifecycle hooks for facade

Dushusir:
为什么要新增 lifecycle 钩子？
注册 canvas 扩展的 Facade API，必须在渲染完成之后才能调用，可以抛出生命周期的钩子，让用户知道已经渲染完成，再调用 canvas 扩展的API

文档待补充

## Pull Request Checklist

### DIFF
--- a/packages/facade/src/apis/facade.ts
+++ b/packages/facade/src/apis/facade.ts
@@ -41,6 +41,7 @@ import { SetFormulaCalculationStartMutation } from '@univerjs/engine-formula';
 import { FDocument } from './docs/f-document';
 import { FWorkbook } from './sheets/f-workbook';
 import { FSheetHooks } from './sheets/f-sheet-hooks';
+import { FHooks } from './f-hooks';
 
 export class FUniver {
     /**
@@ -323,6 +324,13 @@ export class FUniver {
      */
     getSheetHooks() {
         return this._injector.createInstance(FSheetHooks);
+    }
+
+    /**
+     * Get hooks
+     */
+    getHooks() {
+        return this._injector.createInstance(FHooks);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


packages/facade/src/apis/facade.ts
```diff
@@ -325,6 +326,13 @@ export class FUniver {
         return this._injector.createInstance(FSheetHooks);
     }
 
+    /**
+     * Get hooks
+     */
+    getHooks() {
```
wzhudev, line 332: The name should be more specific, like `getLifecycleHooks`. Unless you're going to bind other types of hooks o `f-hook`. (And what are other possible books?)

Dushusir: I have considered this, and it is indeed worth discussing.

The current design is divided into 4 types of hooks, FHooks general hooks, FSheetHooks/FDocHooks/FSlideHooks hooks for specific document types

Although the role of hooks can be replaced by listening to Command in most scenarios, hooks are more semantic and more popular. In the future, we may refer to hansontable to add some hooks https://handsontable.com/docs/javascript-data-grid/api/hooks/
Add afterDispose, afterCopy, afterPaste... to FHooks

This brings up a management issue. Should the internal distinction of the 4 types of hooks be refined?
If it is necessary to refine, it can be divided into
FLifecycleHooks: including life cycle, dispose event
FClipboardHooks: clipboard hook
FRowHooks
FColumnHooks
FCellHooks
etc.

Which solution is better?

Aga92-92 (approved):

packages/facade/src/apis/facade.ts
wzhudev: OK. Let's merge all hooks in this class. Maybe it's better for fresh users to get what they want.

wzhudev (approved):
LGTM

PR #2375, opened 2024-05-31
feat: permission remaining issues

ybzky:
close #xxx

## Pull Request Checklist


packages/core/src/types/interfaces/i-cell-data.ts
```diff
@@ -73,6 +73,7 @@ export interface ICellMarks {
     tr?: ICellMarksStyle;
     bl?: ICellMarksStyle;
     br?: ICellMarksStyle;
+    isSkip?: boolean;
```
wzhudev, line 76: This property appears in 6 different places. Not smelling good. What does it mean?

ybzky: use to skip extension render; each extension manages its own rendering logic

ybzky: create a ticket optimization later and put permission on the top layer to caculate @ybzky

wzhudev (approved):
LGTM

PR #2377, opened 2024-05-31
fix(engine-render): cell text box canvas cannot render when canvas width or height is 0

lirongrong-github:
Before:

DOMException: Failed to execute 'drawImage' on 'CanvasRenderingContext2D': The image argument is a canvas element with a width or height of 0.

I can't reproduce it, but it did happen locally. I tried to start a new project and it worked fine, deleting node_modules and caching and rolling back older versions didn't solve the problem

After: -->

cell input render normally


packages/engine-render/src/context.ts
```diff
@@ -571,6 +571,10 @@ export class UniverRenderingContext2D implements CanvasRenderingContext2D {
 
         const _context = this._context;
 
+        if (!this._isDrawable) {
```
Jocs, line 574: _isDrawable is a function, but you used as a property visit.

packages/engine-render/src/context.ts
```diff
@@ -580,6 +584,22 @@ export class UniverRenderingContext2D implements CanvasRenderingContext2D {
         }
     }
 
+    _isDrawable(args: any[]): boolean {
```
Jocs, line 587: If it is a private method, add a `private`?

PR #2734, opened 2024-07-10
fix(sheets-ui): bugfix for data-validation/comments

weird94:
close #xxx

## Pull Request Checklist


packages/core/src/common/url.ts
```diff
@@ -14,9 +14,47 @@
  * limitations under the License.
  */
 
-const expression = /[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)/gi;
-const regex = new RegExp(expression);
-
+const re_weburl = new RegExp(
```
wzhudev, line 17: Should have test cases to cover this regular expression. And regular expression are **values with state**. It is recommended to create different instances instead of reuse the same one (or reset its state before another calling) to avoid potential problems.

packages/core/src/shared/tools.ts
```diff
@@ -447,6 +447,10 @@ export class Tools {
         return isLegalUrl(url);
     }
 
+    static normalizeUrl(url: string) {
```
wzhudev, line 450: Do not add more methods to the `Tools` class. It is considered a technical dept to wrap util functions in a class.

packages/core/src/types/interfaces/i-document-data.ts
```diff
@@ -130,6 +130,11 @@ export interface IDocumentBody {
 
     customRanges?: ICustomRange[]; // plugin register，implement special logic for streams， hyperlink, field，structured document tags， bookmark，comment
     customDecorations?: ICustomDecoration[];
+
+    /**
+     * for copy/paste, data of custom-range and other module
+     */
+    payloads?: Record<string, string>;
```
wzhudev, line 137: It would be saved to the disk?

packages/sheets-thread-comment/src/controllers/sheets-thread-comment.controller.ts
```diff
@@ -45,6 +45,8 @@ export const DefaultSheetsThreadCommentConfig: IUniverSheetsThreadCommentConfig
 
 @OnLifecycle(LifecycleStages.Starting, SheetsThreadCommentController)
 export class SheetsThreadCommentController extends Disposable {
+    private _setting = false;
```
wzhudev, line 48: What is the flag used for? It deservers comments.

packages/sheets-hyper-link-ui/src/controllers/set-range.controller.ts
```diff
@@ -14,23 +14,34 @@
...
 import { Inject, Injector } from '@wendellhu/redi';
 import type { ISetRangeValuesMutationParams } from '@univerjs/sheets';
 import { ClearSelectionAllCommand, ClearSelectionContentCommand, ClearSelectionFormatCommand, getSheetCommandTarget, SetRangeValuesCommand, SetRangeValuesMutation, SetRangeValuesUndoMutationFactory, SheetInterceptorService, SheetsSelectionsService } from '@univerjs/sheets';
 import type { IAddHyperLinkCommandParams, IUpdateHyperLinkCommandParams } from '@univerjs/sheets-hyper-link';
 import { AddHyperLinkCommand, AddHyperLinkMutation, HyperLinkModel, RemoveHyperLinkMutation, UpdateHyperLinkCommand } from '@univerjs/sheets-hyper-link';
+import { IEditorBridgeService } from '@univerjs/sheets-ui';
+import { DOC_HYPER_LINK_PLUGIN } from '@univerjs/docs-hyper-link';
 import { isLegalLink, serializeUrl } from '../common/util';
 
+const getPlainTextFormRich = (data: IDocumentData) => {
```
wzhudev, line 28: This method should be provided in `docs` package and should have test covered.

packages/sheets-thread-comment/src/controllers/sheets-thread-comment.controller.ts
weird94: Changed to `_isSwitchToCommenting` means is switching to an comment cell directly(maybe switch to a sub-sheet), avoid to be effect by selection init of sub-sheet if it was not initialized before

packages/core/src/types/interfaces/i-document-data.ts
weird94: add an comment to mark it should not saved to disk

wzhudev (approved):
LGTM

PR #2766, opened 2024-07-13
feat: add telemetry to univer and capture rendering metrics

lumixraku:
close #xxx

## Pull Request Checklist


packages/core/src/shared/tools.ts
```diff
@@ -704,4 +704,12 @@ export class Tools {
     static clamp(value: number, min: number, max: number) {
         return Math.max(min, Math.min(max, value));
     }
+
+    static Now(): number {
```
jikkai, line 708: Why does the first letter of the static member method name need to be capitalized❓

PR #2788, opened 2024-07-16
fix(design): add class name as poperty of select list

VicKun4937:
- add some optional property  for custom css


packages/design/src/components/select-list/SelectList.tsx
```diff
@@ -48,10 +48,20 @@ export interface ISelectListProps {
...
+
+    /**
+     * The className each option
+     */
+    optionClassName?: string;
+
+    /**
+     * The className of the wrapper
+     */
+    wrapperClassName?: string;
```
jikkai, line 60: Should be named `className`

jikkai (changes requested):

jikkai (approved):
LGTM

PR #2882, opened 2024-07-30
fix(sheet): fix expand range performance

VicKun4937:
1.  fix sheet expand performance

the reason is we copy the the Matrix  too many times, every cusor change we copy a Matrix  , for this problem , just use the all sheet cell matrix instead , we just change the range property


wzhudev (commented):
LGTM! Good job.

wzhudev (approved):
LGTM! Good job.

PR #3027, opened 2024-08-12
feat(sheet): matrix test

VicKun4937:
- add  test for getdatarange


VicKun4937:
image
 the function is not performance enough, so , I think we should keep both function

DR-Univer (approved):

Issue #1803, opened 2024-04-04
[Feature] Change Default Currency Symbol


joanengenhoff:
### Initial checklist

- [X] Is this really a problem?
- [X] I have searched the Github Issues for similar issues, but did not find anything.

### Problem

I want to use multiple currency symbols. And make Euro as default (in the toolbar and inside the sheets rows), but I haven't found a way to do that in the docs or in the settings object.

How can I achive that? If it's possible.

### Solution

Be able to change the currency icon in the UI config object.

Gggpound:
Certainly, if you need to add shortcut currency buttons, you can modify the code in univer/packages/sheets-numfmt/src/menu/menu.ts. This file is responsible for handling the menu functionality related to number formatting in the application.

Matt-Jensen:
> Certainly, if you need to add shortcut currency buttons, you can modify the code in univer/packages/sheets-numfmt/src/menu/menu.ts. This file is responsible for handling the menu functionality related to number formatting in the application.

Is this the currently supported route for changing the default currency?
[Extending UI](https://univer.ai/guides/sheet/customization/ui.en-US#how-to-register-custom-menu-in-third-party-plugin).  Then we'd need to implement our own plugin, controller(s), command handler(s), and icon registration?  However we'd also need to disable the default RMB currency menu items and replace them first?

PR #3062, opened 2024-08-14
feat(numfmt): currency shortcuts support internationalization

Gggpound:
close https://github.com/dream-num/univer/issues/1803

This time supports quick keys for USD, RMB, and EUR. Other currencies will be added after internationalization.

## Pull Request Checklist

### DIFF
--- a/packages/sheets-numfmt/src/base/const/CURRENCY-SYMBOLS.ts
+++ b/packages/sheets-numfmt/src/base/const/CURRENCY-SYMBOLS.ts
@@ -43,3 +43,20 @@ export const currencySymbols = [
     '₾',
     '₿',
 ];
+
+export const countryCurrencyMap = {
+    US: '$', // United States Dollar
+    CA: 'C$', // Canadian Dollar
+    GB: '£', // British Pound Sterling
+    JP: '¥', // Japanese Yen
+    IN: '₹', // Indian Rupee
+    AU: 'A$', // Australian Dollar
+    CN: '¥', // Chinese Yuan
+    KR: '₩', // South Korean Won
+    RU: '₽', // Russian Ruble
+    // Euro countries
+    AT: '€', BE: '€', CY: '€', EE: '€', FI: '€', FR: '€',
+    DE: '€', GR: '€', IE: '€', IT: '€', LV: '€', LT: '€',
+    LU: '€', MT: '€', NL: '€', PT: '€', SK: '€', SI: '€', ES: '€',
+    // Add more mappings as needed
+};
--- a/packages/sheets-numfmt/src/commands/commands/set-currency.command.ts
+++ b/packages/sheets-numfmt/src/commands/commands/set-currency.command.ts
@@ -19,6 +19,8 @@ import { CommandType, ICommandService, Range } from '@univerjs/core';
 import { SheetsSelectionsService } from '@univerjs/sheets';
 
 import { CURRENCYFORMAT } from '../../base/const/FORMATDETAIL';
+import { countryCurrencyMap } from '../../base/const/CURRENCY-SYMBOLS';
+import { NumfmtMenuController } from '../../controllers/numfmt.menu.controller';
 import type { ISetNumfmtCommandParams } from './set-numfmt.command';
 import { SetNumfmtCommand } from './set-numfmt.command';
 
@@ -28,14 +30,15 @@ export const SetCurrencyCommand: ICommand = {
     handler: async (accessor: IAccessor) => {
         const commandService = accessor.get(ICommandService);
         const selectionManagerService = accessor.get(SheetsSelectionsService);
-
+        const numfmtMenuController = accessor.get(NumfmtMenuController);
+        const symbol = countryCurrencyMap[numfmtMenuController.getCurrencySymbol()] || '$';
         const selections = selectionManagerService.getCurrentSelections();
         if (!selections || !selections.length) {
             return false;
         }
         const values: ISetNumfmtCommandParams['values'] = [];
 
-        const suffix = CURRENCYFORMAT[0].suffix('¥');
+        const suffix = CURRENCYFORMAT[0].suffix(symbol);
 
         selections.forEach((selection) => {
             Range.foreach(selection.range, (row, col) => {
--- a/packages/sheets-numfmt/src/controllers/numfmt.menu.controller.ts
+++ b/packages/sheets-numfmt/src/controllers/numfmt.menu.controller.ts
@@ -18,8 +18,10 @@ import { Disposable, Inject, Injector, LifecycleStages, OnLifecycle } from '@uni
 import type { MenuConfig } from '@univerjs/ui';
 import { ComponentManager, IMenuService } from '@univerjs/ui';
 
+import { BehaviorSubject } from 'rxjs';
 import { AddDecimalMenuItem, CurrencyMenuItem, FactoryOtherMenuItem, PercentMenuItem, SubtractDecimalMenuItem } from '../menu/menu';
 import { MORE_NUMFMT_TYPE_KEY, MoreNumfmtType, Options, OPTIONS_KEY } from '../components/more-numfmt-type/MoreNumfmtType';
+import type { countryCurrencyMap } from '../base/const/CURRENCY-SYMBOLS';
 
 export interface IUniverSheetsNumfmtConfig {
     menu: MenuConfig;
@@ -29,6 +31,9 @@ export const DefaultSheetNumfmtConfig = {};
 
 @OnLifecycle(LifecycleStages.Rendered, NumfmtMenuController)
 export class NumfmtMenuController extends Disposable {
+    private _currencySymbol$ = new BehaviorSubject<keyof typeof countryCurrencyMap>('US');
+    public readonly currencySymbol$ = this._currencySymbol$.asObservable();
+
     constructor(
         private readonly _config: Partial<IUniverSheetsNumfmtConfig>,
         @Inject(Injector) private _injector: Injector,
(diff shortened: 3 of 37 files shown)


Jocs (approved):

packages/sheets-numfmt/src/base/const/CURRENCY-SYMBOLS.ts
```diff
@@ -43,3 +45,21 @@ export const currencySymbols = [
...
+    IN: '₹', // Indian Rupee
+    AU: 'A$', // Australian Dollar
+    [LocaleType.ZH_CN]: '¥', // Chinese Yuan
+    [LocaleType.ZH_TW]: '¥', // Chinese Yuan
+    KR: '₩', // South Korean Won
+    [LocaleType.RU_RU]: '₽', // Russian Ruble
+    // Euro countries
+    AT: '€', BE: '€', CY: '€', EE: '€', FI: '€', FR: '€',
+    DE: '€', GR: '€', IE: '€', IT: '€', LV: '€', LT: '€',
+    LU: '€', MT: '€', NL: '€', PT: '€', SK: '€', SI: '€', ES: '€',
```
lovasoa, line 63: Is it normal that two-letters country codes are mixed with full locale strings like `en-US` ?

packages/sheets-numfmt/src/commands/commands/set-currency.command.ts
```diff
@@ -28,14 +29,17 @@ export const SetCurrencyCommand: ICommand = {
     handler: async (accessor: IAccessor) => {
         const commandService = accessor.get(ICommandService);
         const selectionManagerService = accessor.get(SheetsSelectionsService);
+        const localeService = accessor.get(LocaleService);
+        const locale = localeService.getCurrentLocale();
```
lovasoa, line 33: Is this going to work for euros ? 
Or will we have to wait until univer is translated into every european language ?
I liked using the browser's current regional settings, because it allowed this to work with languages univer is not (and will never be) translated into.

packages/sheets-numfmt/src/base/const/CURRENCY-SYMBOLS.ts
Gggpound: Here, currency types are represented using language codes instead of coutry codes (You can find specific language codes in the file located at (packages/core/src/types/enum/locale-type.ts)[https://github.com/dream-num/univer/blob/dev/packages/core/src/types/enum/locale-type.ts]).

packages/sheets-numfmt/src/commands/commands/set-currency.command.ts
Gggpound: No, no, no. The European region is also a future direction for our services. It's just that we currently don't have enough manpower to support this work. Most of the internationalization resources in the project are provided by the community, and we are very grateful for their help.
Can we try adding an enumeration of LocaleType euro, but translate the text in English (Wait for the open source community to provide the corresponding language copywriting)?

Gggpound: If you want to implement this button in the toolbar by yourself, you can try overriding this menu configuration.
After loading the numfmt-plugin, you can reset the(CurrencyMenuItem)[packages/sheets-numfmt/src/menu/menu.ts] by calling the(setMenuItem)[packages/ui/src/services/menu/menu.service.ts] method in menu.service.ts.

lovasoa: The thing is, there are many many different languages in the "euro region". What would an euro locale contain?

lovasoa: Are you sure you don't want to decouple supported languages from supported currencies?

Gggpound: We haven't thought about this for a while, but from a design point of view,
the current language isn't platform-specific, it's available through the LocaleService in project

Gggpound: 【there are many many different languages in the "euro region"】Could there be multiple languages for the same currency? That means that our localType enumerations may be excessive.
Please give me a little more time to discuss it internally

lovasoa: Ideally, I would like the default currency displayed to SQLPage users to be the currency of the place they are in, even if the rest of the UI is in English. 

There are 20 countries in the eurozone, speaking 19 languages. Some of these languages have <1M locutors. I understand you have grand ambitions, but univerjs will probably not be translated in all of these languages anytime soon. It would be great if we could still have the € sign by default...

Gggpound: Sure, I think providing an API to set a custom currency symbol is a great solution. It allows you to set the currency symbol independently of the platform language and `Univer` own language settings. When integrating with `Univer`, you can use the API to set the currency symbol, regardless of how you obtain the system language.

lovasoa: I think there is a misunderstanding. I don't need an api to set the default currency symbol, I would like the default symbol to automatically reflect the users preference, wherever they are. That's what my initial pull request did.

packages/sheets-numfmt/src/base/const/CURRENCY-SYMBOLS.ts
lovasoa: Yes, and that's a problem, because the currency depends on the country, but not on the language

packages/sheets-numfmt/src/commands/commands/set-currency.command.ts
Gggpound: 1. We will provide an API for setting the currency symbol, and you can write some code to call it. For example, you can use the browser's user-agent to recognize the country code and set the corresponding symbol.

2. We will add a country code identifier to the snapshot upon creation. If not set, it will default to the mapped value of that country code and will not follow the language settings.
 
3. Later, we will provide a StackBlitz code to guide you through the implementation.

There are some blocking bugs that need to be resolved, and we expect to provide them to you by next week.

packages/sheets-numfmt/src/controllers/numfmt.menu.controller.ts
```diff
@@ -29,6 +31,8 @@ export const DefaultSheetNumfmtConfig = {};
 
 @OnLifecycle(LifecycleStages.Rendered, NumfmtMenuController)
 export class NumfmtMenuController extends Disposable {
+    private _currencySymbol$ = new BehaviorSubject<keyof typeof currencyMap>('ZH_CN');
```
wzhudev, line 34: It is better to infer the default currency symbol for the browser's `navgiator` info. Since this property is at UI layer, it is considered as harmless.

wzhudev: It it recommended to make property `readonly` whenever possible.

packages/sheets-numfmt/src/controllers/numfmt.menu.controller.ts
```diff
@@ -29,6 +31,8 @@ export const DefaultSheetNumfmtConfig = {};
...
+    public currencySymbol$ = this._currencySymbol$.asObservable();
```
wzhudev, line 35: Add a blank line before constructors to make the code clear to read.

packages/sheets-numfmt/src/index.ts
```diff
@@ -16,6 +16,7 @@
 
 export { UniverSheetsNumfmtPlugin } from './numfmt-plugin';
 export { getPatternPreview } from './utils/pattern';
+export { NumfmtMenuController } from './controllers/numfmt.menu.controller';
```
wzhudev, line 19: Currency symbols should be exported as well so users know what currencies are available.

packages/sheets-numfmt/src/controllers/numfmt.menu.controller.ts
```diff
@@ -50,4 +54,12 @@ export class NumfmtMenuController extends Disposable {
         this.disposeWithMe((this._componentManager.register(MORE_NUMFMT_TYPE_KEY, MoreNumfmtType)));
         this.disposeWithMe((this._componentManager.register(OPTIONS_KEY, Options)));
     }
+
+    public setCurrencySymbol(symbol: keyof typeof currencyMap) {
```
wzhudev, line 58: JSDoc is expected to be required on this method if it is intended to be used directly by users.

wzhudev (changes requested):
Overall looks good to me. But there are some code quality issues.

Gggpound:
After the Univer instance is created, execute the following code
```ts
import { NumfmtMenuController } from '@univerjs/sheets-numfmt';

const univer = new Univer();

univer.registerPlugin(UniverSheetsNumfmtPlugin);

setCurrencySymbol()
function setCurrencySymbol() {
    const injector = univer.__getInjector();
    const lifecycleService = injector.get(LifecycleService);

    function calculateYourCurrencySymbol() {
        const numfmtMenuController = injector.get(NumfmtMenuController);
        numfmtMenuController.setCurrencySymbol('RU');
    }

    const dispose = lifecycleService.lifecycle$.subscribe((stage) => {
        if (LifecycleStages.Rendered === stage) {
            calculateYourCurrencySymbol();
            dispose.unsubscribe();
        }
    });
}

```

wzhudev (approved):
LGTM. Please add an example or some documentation on univer.ai to demonstrate how this works. @Gggpound

thieu-hoa:
Hello,

I am currently using UniverJS version 5.5, and I 'm experiencing issues with setting the currency symbol in the toolbar. Despite following previous suggestions, I still haven't been able to change the currency symbol to "€" (Euro) instead of the "$" or the default "¥" . My locale settings is : "locale: LocaleType.FR_FR". I have also adjusted the imports as follows: import DesignFrFR from '@univerjs/design/locale/fr-FR';
import UIFrFR from '@univerjs/ui/locale/fr-FR';
import DocsUIFrFR from '@univerjs/docs-ui/locale/fr-FR';
import SheetsFrFR from '@univerjs/sheets/locale/fr-FR';
import SheetsUIFrFR from '@univerjs/sheets-ui/locale/fr-FR';
import SheetsFormulaUIFrFR from '@univerjs/sheets-formula-ui/locale/fr-FR';
import SheetsNumfmtUIFrFR from '@univerjs/sheets-numfmt-ui/locale/fr-FR';

Is there any official documentation or updated example that demonstrates how to configure this correctly?

Thank you in advance for your help !

PR #3158, opened 2024-08-22
feat(facade): add unit hooks

hexf00:
Hooks naming will be standardized to the format: prefix + verb in present tense + noun.

- [x] beforeCreateUnit
- [x] afterCreateUnit
- [x] beforeDisposeUnit
- [x] afterDisposeUnit

## Pull Request Checklist

### DIFF
--- a/packages/core/src/services/instance/instance.service.ts
+++ b/packages/core/src/services/instance/instance.service.ts
@@ -47,6 +47,11 @@ export interface ICreateUnitOptions {
  * It also manages the focused univer instance.
  */
 export interface IUniverInstanceService {
+    /** A set of callbacks that will be called before a new unit is added. */
+    beforeUnitAddCallbacks: Set<(unit: UnitModel) => boolean | void>;
+    /** A set of callbacks that will be called before a unit is disposed. */
+    beforeUnitDisposeCallbacks: Set<(unit: UnitModel) => boolean | void>;
+
     /** Omits value when a new UnitModel is created. */
     unitAdded$: Observable<UnitModel>;
     /** Subscribe to curtain type of units' creation. */
@@ -104,10 +109,16 @@ export class UniverInstanceService extends Disposable implements IUniverInstance
         super();
     }
 
+    beforeUnitDisposeCallbacks = new Set<(unit: UnitModel) => boolean | void>();
+
+    beforeUnitAddCallbacks = new Set<(unit: UnitModel) => boolean | void>();
+
     override dispose(): void {
         super.dispose();
 
         this._focused$.complete();
+        this.beforeUnitAddCallbacks.clear();
+        this.beforeUnitDisposeCallbacks.clear();
     }
 
     private _createHandler!: (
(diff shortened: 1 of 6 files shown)


packages/core/src/services/instance/instance.service.ts
```diff
@@ -172,6 +183,15 @@ export class UniverInstanceService extends Disposable implements IUniverInstance
     __addUnit(unit: UnitModel, options?: ICreateUnitOptions): void {
         const type = unit.type;
 
+        if (this.beforeUnitAddCallbacks.size > 0) {
+            for (const callback of this.beforeUnitAddCallbacks) {
+                if (callback(unit) === false) {
+                    // if not use throw error, the createUnit returns null
+                    throw new Error('[UniverInstanceService]: cannot add unit with user callback.');
```
hexf00, line 190: @wzhudev Check if this has been handled properly.

packages/facade/src/apis/__tests__/f-hooks.spec.ts
```diff
@@ -50,3 +50,86 @@ describe('Test FUniver', () => {
...
+            return false;
+        });
+        univerAPI.getHooks().afterCreateUnit(disposeCallback);
+
+        const unit = univer.createUnit(UniverInstanceType.UNIVER_SHEET, {
+            id: randomUnitId,
+        });
+
+        // TODO: not calling dispose() method
+        // unit.dispose();
```
hexf00, line 117: @wzhudev The unit.dispose() function is not associated; in the future, a better and more unified mechanism is needed. Consider moving `dispose` to `UnitModel` or an intermediate layer. Use `dispose` for the destroy function.

packages/core/src/services/instance/instance.service.ts
wzhudev: I don't know. It is your responsibility to check if it is handled properly.

packages/core/src/services/instance/instance.service.ts
```diff
@@ -47,6 +47,11 @@ export interface ICreateUnitOptions {
  * It also manages the focused univer instance.
  */
 export interface IUniverInstanceService {
+    /** A set of callbacks that will be called before a new unit is added. */
```
wzhudev, line 50: Should never expose internal properties as mutable. Instead, provide methods to do that.

packages/facade/src/apis/__tests__/f-hooks.spec.ts
```diff
@@ -50,3 +50,86 @@ describe('Test FUniver', () => {
...
+
+        const _univerInstanceService = get(IUniverInstanceService);
```
wzhudev, line 119: Not necessary to add an underscore before a variable. The underbar has a special usage for "not used variables".

packages/facade/src/apis/__tests__/f-hooks.spec.ts
wzhudev: Already clearified this with hexf00 offline.

wzhudev (changes requested):

PR #3187, opened 2024-08-26
feat(sheet): performan the expand range

VicKun4937:
-  if the sheet has none span, we can ignore span check logic
- we treat the span merge cells has same value just like the [left, top] cell
- we should check the span which may be added in the most left , top, right, bottom, so that can avoid the part of span be added in result range.

# Here is performance compare:
the count is how many times we called ` matrix.getValue(row, col)`
before:
Image
after with span check:
Image
without checked span:
Image


wzhudev (approved):
LGTM

Issue #3089, opened 2024-08-16
[Bug] Import Excel, freezes after scrolling


zhanghui6666:
### Before you submit this issue, have you checked the following

- [X] Is this really a problem?
- [X] I have searched the Github Issues for similar issues, but did not find anything.

### Affected packages and versions

0.2.6

### Reproduction link

[电镀日产量报表_1723597532980.xlsx](https://github.com/user-attachments/files/16633112/_1723597532980.xlsx)

### Expected behavior

正常滚动

### Actual behavior

我在后端生成了excel并转换为base64，然后在前端再将base64转换为excel的file文件，调用了importXLSXToSnapshot后成功渲染了，但是在滚动后就卡死了，这是什么情况
I generated an Excel file on the backend and converted it to base64. Then, I converted the base64 file to an Excel file on the frontend and successfully rendered it by calling importXLSXoSnapshot. However, it froze after scrolling. What is the situation

lumixraku:
a little confused, I updated your file to univer space, it rendered very smooth.
How to reproduce this problem?
<img width="832" alt="image" src="https://github.com/user-attachments/assets/d89b9ef5-cc73-4dd6-baa5-e508498ac328">

Issue #2968, opened 2024-08-08
Horizontal scroll axis appears even though the columns of the table are not filled up


maotong06:
### Before you submit this issue, have you checked the following

- [X] Is this really a problem?
- [X] I have searched the Github Issues for similar issues, but did not find anything.

### Affected packages and versions

0.2.6

### Reproduction link

https://univer.ai/zh-CN/examples/sheets
in the hyperlink sheet
![image](https://github.com/user-attachments/assets/4b9f48a9-c5a6-4b8c-9ce2-6987032f6f10)

### Expected behavior

Hopefully no scroll axis when the front column just fills up the space

### Actual behavior

Hopefully no scroll axis when the front column just fills up the space

jikkai:
Duplicate of #2836

Issue #2836, opened 2024-07-22
[Bug] Create a new univer table, the scroll wheel does not dynamically change to adapt to the table content


ForSakuraLove:
### Before you submit this issue, have you checked the following

- [X] Is this really a problem?
- [X] I have searched the Github Issues for similar issues, but did not find anything.

### Affected packages and versions

0.2.4

### Reproduction link

The import example on the official website, I used it to import Excel, the scroll wheel did not adapt, I needed to change the row height, then the scroll wheel would adapt. My own creation is the same, so I won't show it.
![image](https://github.com/user-attachments/assets/fb667c49-2474-4a7f-a6dd-da95684aa593)

### Expected behavior

Scroll wheel requires adaptive content

### Actual behavior

Scroll wheel has no adaptive content

### System information

System:
    OS: Windows 10 10.0.19045
    CPU: (6) x64 Intel(R) Core(TM) i5-9500 CPU @ 3.00GHz
    Memory: 5.89 GB / 19.79 GB
  Binaries:
    Node: 18.20.0 - D:\Program Files\nodejs\node.EXE
    npm: 10.5.0 - D:\Program Files\nodejs\npm.CMD
    pnpm: 9.5.0 - D:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Edge: Chromium (126.0.2592.113)
    Internet Explorer: 11.0.19041.4355

lumixraku:
@ForSakuraLove Is this issue still present? I was unable to reproduce it.

I am uploading this excel file.
[manycol.xlsx](https://github.com/user-attachments/files/16621365/manycol.xlsx)

ForSakuraLove:
@lumixraku This problem still exists. I use the Excel file you uploaded, but it still doesn't show all
![image](https://github.com/user-attachments/assets/f0dcc4c6-8877-437e-aa43-d954d433ab75)

MurphyHou123:
I have the same issue now. the maximum show rows are 42. actually the totall rows are 1000.
![image](https://github.com/user-attachments/assets/25668443-0821-4378-b0f2-267d93513ad3)

PR #3205, opened 2024-08-27
fix: sheet dispose cause scroll update scene size failed.

lumixraku:
close https://github.com/dream-num/univer/issues/2836
close https://github.com/dream-num/univer/issues/3089
close https://github.com/dream-num/univer/issues/2968

## What Exactly Happened?
This issue occurred in the business.univer.ai import and export example. There is no such problem when importing an Excel file into the univer space.

It appears that the scrollbar is not functioning. In fact, this is due to the scenemap being null after disposal. However, when a univer is recreated, the scrollbar is initialized before the scenemap value is set, and it requires the width and height of the scene. If the scene is null, the scrollbar defaults to a width of 1500.

The actual problem arises when recreating the univer shortly after disposal, for example, within 2000 milliseconds.

## Why is there no problem on the first creation of the Univer?
The ScrollRenderController is registered during the rendering stage. The scenemap has its value set at this stage.
But when recreating the univer after disposal, all dependencies have already been registered, so the ScrollRenderController is initialized before the scenemap sets its value.
 

## How to reproduce this issue in univer? 

in univer/examples/src/sheets/main.ts
add this after univer.createUnit(xxx)
```
setTimeout(async () => {
    window.univerAPI.disposeUnit('workbook-01');

    await setTimeout(() => { }, 1000);
    univer.createUnit(UniverInstanceType.UNIVER_SHEET, DEFAULT_WORKBOOK_DATA_DEMO);
}, 2000);
```

This issue can be reproduced in business.univer.ai  import & export example.

## Pull Request Checklist

### DIFF
--- a/examples/src/sheets/main.ts
+++ b/examples/src/sheets/main.ts
@@ -145,12 +145,31 @@ if (!IS_E2E) {
 setTimeout(() => {
     import('./lazy').then((lazy) => {
         const plugins = lazy.default();
-        plugins.forEach((p) => univer.registerPlugin(p[0], p[1]));
+        plugins.forEach((p) => {
+            univer.registerPlugin(p[0], p[1]);
+        });
     });
 }, LOAD_LAZY_PLUGINS_TIMEOUT);
 
 window.univer = univer;
 window.univerAPI = FUniver.newAPI(univer);
+
+setTimeout(async () => {
+    window.univerAPI.disposeUnit('workbook-01');
+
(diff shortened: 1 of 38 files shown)


lumixraku:
WeChatWorkScreenshot_d9aba3d4-2cd0-440d-9e3e-bc32ac6f988e

wzhudev (approved):
LGTM

PR #3230, opened 2024-08-28
fix(sheets-drawing): drawing snapshot error

weird94:
close #xxx

## Pull Request Checklist

### DIFF
--- a/packages/sheets-drawing/src/controllers/sheet-drawing.controller.ts
+++ b/packages/sheets-drawing/src/controllers/sheet-drawing.controller.ts
@@ -66,6 +66,7 @@ export class SheetsDrawingLoadController extends Disposable {
                 },
                 onLoad: (unitId, value) => {
                     this._sheetDrawingService.registerDrawingData(unitId, value);
+                    this._drawingManagerService.registerDrawingData(unitId, value);
                 },
             })
         );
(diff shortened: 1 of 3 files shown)


packages/sheets-drawing/src/controllers/sheet-drawing.controller.ts
```diff
@@ -66,6 +66,7 @@ export class SheetsDrawingLoadController extends Disposable {
                 },
                 onLoad: (unitId, value) => {
                     this._sheetDrawingService.registerDrawingData(unitId, value);
+                    this._drawingManagerService.registerDrawingData(unitId, value);
```
wzhudev, line 69: Why should we register the value twice on different service? IMHO, `_sheetDrawingService` should do this.

weird94: I don't know why, but this was actually 2 service to manager drawing

weird94: > Why should we register the value twice on different service? IMHO, `_sheetDrawingService` should do this.

I don't know why, but this was actually 2 service to manager drawing

weird94: cc @DR-Univer

Jocs: Doc also has two services, but docDrawingService will be deprecated in the future.

wzhudev (approved):
LGTM

PR #3303, opened 2024-09-03
fix(sheet): fix formular performance

VicKun4937:
- performance issue


VicKun4937:
image
the  used  mom in  a bad case

PR #3415, opened 2024-09-12
fix: the viewport size is too big after cancel filter rule, then viewport main is blank because graphic mem is full.

lumixraku:
close https://github.com/dream-num/univer-pro/issues/2479

## Pull Request Checklist

### DIFF
--- a/packages/engine-render/src/viewport.ts
+++ b/packages/engine-render/src/viewport.ts
@@ -47,8 +47,8 @@ interface IViewProps extends IViewPosition {
     isWheelPreventDefaultY?: boolean;
     active?: boolean;
 
-    isRelativeX?: boolean;
-    isRelativeY?: boolean;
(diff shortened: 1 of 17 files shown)


packages/engine-render/src/viewport.ts
```diff
@@ -194,9 +194,12 @@ export class Viewport {
 
     private _paddingEndY: number = 0;
 
-    private _isRelativeX: boolean = false;
-
-    private _isRelativeY: boolean = false;
+    private _explicitViewportWidthSet: boolean = false;
```
Jocs, line 197: 这个属性要不要加一个注释？

Jocs (approved):

packages/engine-render/src/viewport.ts
lumixraku: 可以， 确实从变量名不好说明

PR #3443, opened 2024-09-14
fix(span): fix editor position not right issue

VicKun4937:
- fix (https://github.com/dream-num/univer-pro/issues/2482)


wzhudev (approved):
LGTM

PR #3489, opened 2024-09-19
feat(sheet): support sheet default style

### LINKED ISSUE #3021: [Feature] I hope the cell can be set to WrapStrategy.CLIP by default
### 初始清单

- [X] 这真的是个问题吗？
- [X] 我已经在 Github Issues 中搜索过了，但没有找到类似的问题。

### 问题

我可以在那里去修改

### 解决方案

希望有一个入口方便我去修改掉默认值

VicKun4937:
close #3021

- support sheet default style

### DIFF
--- a/packages/core/src/shared/tools.ts
+++ b/packages/core/src/shared/tools.ts
@@ -744,7 +744,7 @@ interface IStyleDataObject {
 /**
  * compose styles by priority, the latter will overwrite the former
  * @param { Nullable<IStyleData>[]} styles the styles to be composed
- * @returns  { Nullable<IStyleData>[]}
+ * @returns  { Nullable<IStyleData>[]} Returns the composed style
  */
 export function composeStyles(...styles: Nullable<IStyleData>[]): IStyleData {
     const result: IStyleData = {};
--- a/packages/core/src/shared/tools.ts
+++ b/packages/core/src/shared/tools.ts
@@ -14,11 +14,12 @@
  * limitations under the License.
  */
 
-import { customAlphabet, nanoid } from 'nanoid';
-
 import mergeWith from 'lodash.mergewith';
+
+import { customAlphabet, nanoid } from 'nanoid';
 import { isLegalUrl, normalizeUrl } from '../common/url';
-import type { IKeyValue } from './types';
+import type { IStyleData } from '../types/interfaces';
+import type { IKeyValue, Nullable } from './types';
 
 const rmsPrefix = /^-ms-/;
 const rDashAlpha = /-([a-z])/g;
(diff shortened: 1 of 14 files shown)


packages/core/src/shared/tools.ts
```diff
@@ -735,3 +736,29 @@ export function generateRandomId(n: number = 21, alphabet?: string): string {
...
+interface IStyleDataObject {
+    [key: string]: unknown;
+}
+
+/**
+ * compose styles by priority, the latter will overwrite the former
+ * @param { Nullable<IStyleData>[]} styles the styles to be composed
+ * @returns  { Nullable<IStyleData>[]}
+ */
+export function composeStyles(...styles: Nullable<IStyleData>[]): IStyleData {
```
wzhudev, line 749: It is recommended to add unit tests for these util functions.

packages/sheets/src/index.ts
```diff
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-/* eslint-disable perfectionist/sort-named-exports */
```
wzhudev, line 17: IHMO you should probably not to remove this line, because we have a special section index.ts files to exports all commands.

wzhudev: cc @jikkai please take a look into this.

packages/engine-render/src/components/sheets/sheet-skeleton.ts
```diff
@@ -1782,10 +1790,20 @@ export class SpreadsheetSkeleton extends Skeleton {
             }
         }
```
wzhudev, line 1792: Will this hurt performance?

wzhudev: Please wait for @lumixraku 's rendering e2e test.

packages/core/src/shared/tools.ts
VicKun4937: it will be added before this PR be merged

packages/sheets/src/index.ts
jikkai: This eslint ignore comment explicitly placed to maintain region comments for code folding and organization.

packages/core/src/sheets/worksheet.ts
```diff
@@ -86,9 +87,95 @@ export class Worksheet {
...
+    /**
+     * Get the style of the column.
+     * @param {number} column The column index
+     * @returns {Nullable<IStyleData>|string} The style of the column
+     */
+    getColumnStyle(column: number): Nullable<IStyleData> | string {
+        return this._styles.get(this._snapshot.columnData[column]?.s);
+    }
+
+    getColumnStyleInternal(column: number): Nullable<IStyleData> {
```
Dushusir, line 99: Why are there two methods here? getColumnStyleInternal and getColumnStyle seem to be essentially the same method. The `Styles.get` method also dynamically determines the type of `s`.

packages/core/src/sheets/worksheet.ts
```diff
@@ -86,9 +87,95 @@ export class Worksheet {
         return this._spanModel;
     }
 
+    /**
+     * Get the style of the column.
+     * @param {number} column The column index
+     * @returns {Nullable<IStyleData>|string} The style of the column
+     */
+    getColumnStyle(column: number): Nullable<IStyleData> | string {
```
Dushusir, line 95: Why are there two methods here? `getColumnStyleInternal` and `getColumnStyle` seem to be essentially the same method. The `Styles.get` method also dynamically determines the type of `s`.

packages/core/src/sheets/worksheet.ts
```diff
@@ -86,9 +87,95 @@ export class Worksheet {
...
+        }
+        return style;
+    }
+
+    /**
+     * Set the style of the row.
+     * @param {number} row
+     * @param {string|Nullable<IStyleData>} style The style to be set
+     */
+    setRowStyle(row: number, style: string | Nullable<IStyleData>): void {
```
Dushusir, line 143: Maybe it would be better to move to `RowManager`

packages/core/src/sheets/worksheet.ts
VicKun4937: it not same , the getColumnStyle will return style id, but the internal one which should not public, it useful in sketon for get style data directly

packages/core/src/sheets/worksheet.ts
VicKun4937: it's a good idea, I will do it before this PR merge into dev

PR #3512, opened 2024-09-20
fix(docs): support special emoticons

Jocs:
close #xxx

## Pull Request Checklist

### DIFF
--- a/packages/engine-render/src/components/docs/layout/block/paragraph/language-ruler.ts
+++ b/packages/engine-render/src/components/docs/layout/block/paragraph/language-ruler.ts
@@ -16,10 +16,10 @@
 
 import type { IParagraph } from '@univerjs/core';
 
-import type { ISectionBreakConfig } from '../../../../../basics/interfaces';
 import { EMOJI_REG, hasArabic, hasSpace, hasTibetan, startWithEmoji } from '../../../../../basics/tools';
 import { createSkeletonLetterGlyph, createSkeletonWordGlyph } from '../../model/glyph';
 import { getFontCreateConfig } from '../../tools';
+import type { ISectionBreakConfig } from '../../../../../basics/interfaces';
 import type { DataStreamTreeNode } from '../../../view-model/data-stream-tree-node';
 import type { DocumentViewModel } from '../../../view-model/document-view-model';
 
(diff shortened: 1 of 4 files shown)


packages/engine-render/src/components/docs/layout/block/paragraph/language-ruler.ts
```diff
@@ -35,19 +35,27 @@ export function otherHandler(
...
     let step = 0;
+    let src = charArray;
 
-    for (let i = 0; i < charArray.length; i++) {
-        const newChar = charArray[i];
+    while (src.length) {
+        const char = src.match(/^[\s\S]/gu)?.[0];
 
-        if (hasSpace(newChar) || startWithEmoji(charArray.substring(i))) {
+        if (char == null) {
```
wzhudev, line 43: Remove `==` from our codebase.

wzhudev (commented):
1. IMHO, linebreaking could not rely on DOM so we can add unit tests to it?
2. We should add pixel diff (visual comparison) e2e before merging this to make sure that we can render emoji and many languages correctly.

Jocs:
> We should add pixel diff (visual comparison) e2e before merging this to make sure that we can render emoji and many languages correctly.

It's a new capability, so why wait until after e2e？

packages/engine-render/src/components/docs/layout/block/paragraph/language-ruler.ts
Jocs: Shouldn `== null` be allowed? because the phonetics here are very clear, which is `=== undefined` or `=== null`.

wzhudev: image

Like this. The core idea is to avoid implicit type casting.

wzhudev (approved):
LGTM

packages/engine-render/src/components/docs/layout/block/paragraph/language-ruler.ts
Jocs: 其实我想表达的是，使用 == null,应该就是明确 `=== null || === undefined` 的意思，在其他地方是不建议使用 ==。之前还看了一篇文章，你所不知道的 JavaScript 作者写的，专门还说了这个问题，我找找

PR #3517, opened 2024-09-21
fix(sheet): fix summary performance

VicKun4937:
- fix multi selection calc error
-  fix performance
- the Welford's logic:  https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance, we can use to calc stdev


VicKun4937:
@zhaolixin7 Please pay attention about:
-  hidden row case and mutilp-selection case.
- the big number or small number

yuhongz (approved):
LGTM

wzhudev (approved):

PR #3613, opened 2024-09-28
chore: cleanup package.json

hexf00:
1. fix umd build
2. clean package.json

### DIFF
--- a/packages/umd/scripts/index.mjs
+++ b/packages/umd/scripts/index.mjs
@@ -24,7 +24,7 @@
 import fs from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { merge } from '@univerjs/core';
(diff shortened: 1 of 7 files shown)


packages/umd/scripts/index.mjs
```diff
@@ -24,7 +24,7 @@
 import fs from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { merge } from '@univerjs/core';
+import { merge } from 'lodash-es';
```
jikkai, line 27: why?

jikkai (changes requested):

packages/umd/scripts/index.mjs
hexf00: umd 打包不了，这里直接引入的ts

jikkai (approved):

PR #3614, opened 2024-09-28
fix: format painter for set styles first then merge cell

lumixraku:
close https://github.com/dream-num/univer-pro/issues/2568

## Pull Request Checklist

### DIFF
--- a/packages/sheets-ui/src/controllers/format-painter/format-painter.controller.ts
+++ b/packages/sheets-ui/src/controllers/format-painter/format-painter.controller.ts
@@ -26,10 +26,10 @@ import {
     Tools,
     UniverInstanceType,
 } from '@univerjs/core';
-import { IRenderManagerService } from '@univerjs/engine-render';
 
+import { IRenderManagerService } from '@univerjs/engine-render';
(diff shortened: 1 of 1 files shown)


packages/sheets-ui/src/controllers/format-painter/format-painter.controller.ts
```diff
@@ -121,18 +121,25 @@ export class FormatPainterController extends Disposable {
...
+                    mergedMainCell = cell;
                     merges.push({
                         startRow: mergeInfo.startRow,
                         startColumn: mergeInfo.startColumn,
                         endRow: mergeInfo.endRow,
                         endColumn: mergeInfo.endColumn,
                     });
+                } else if (isMerged) {
+                    //see univer-pro/issues/2568
+                    if (mergedMainCell) {
```
wzhudev, line 140: Could this value be `null`?

wzhudev (approved):

packages/sheets-ui/src/controllers/format-painter/format-painter.controller.ts
lumixraku: It won't.   

PS: we found another issue relatived to this problem. This solution did not cover all case. so closed first.

PR #3629, opened 2024-09-29
fix: prevent delete drawing when input focus

siam-ese:
close [#2628](https://github.com/dream-num/univer-pro/issues/2628)

## Pull Request Checklist


wzhudev (changes requested):
This solution is not suitable for all situations. Let's suppose that if you have other input elements in `DrawingTransform.ts` or elsewhere, you have to edit these components one by one. There should be a generic solution to this problem.

PR #3864, opened 2024-10-26
fix(drawing): fix inner dom position

VicKun4937:
- fix float dom inner size, it should smaller than wrapper


packages/sheets-drawing-ui/src/services/canvas-float-dom-manager.service.ts
```diff
@@ -318,6 +318,10 @@ export class SheetCanvasFloatDomManagerService extends Disposable {
                     if (isChart) {
                         imageConfig.fill = 'white';
                         imageConfig.rotateEnabled = false;
+                        imageConfig.stroke = 'blue';
```
DR-Univer, line 321: 默认blue很丑，来个rgb(216,216,216)

DR-Univer (approved):

PR #3873, opened 2024-10-28
fix(drawing): fix remove drawing elements the control when not remove…

VicKun4937:
- as title , add lisener to remove the control


packages/engine-render/src/scene.transformer.ts
```diff
@@ -1529,6 +1529,22 @@ export class Transformer extends Disposable implements ITransformerConfig {
         this._clearControl$.next(changeSelf);
     }
 
+    /**
+     * @description Clear the control of the object with the specified id
+     * @param {string[]} ids the id of the object to be cleared
+     */
+    public clearControlByIds(ids: string[]) {
+        for (const id of ids) {
+            // const control = this._transformerControlMap.get(id);
```
wzhudev, line 1538: Remove commented code.

wzhudev (approved):

PR #3881, opened 2024-10-29
feat: filter date group

ybzky:
close #xxx

## Pull Request Checklist


packages/design/src/components/tree/Tree.tsx
```diff
@@ -34,54 +35,64 @@ export interface ITreeNodeProps {
     children?: ITreeNodeProps[];
 }
 
+export interface IAttachTreeProps {
+    level?: number;
+    count?: number;
+}
+
+type TreeItemProps = ITreeNodeProps & IAttachTreeProps;
```
wzhudev, line 43: Should start with an `I` as prefix.

packages/sheets-filter-ui/src/locale/en-US.ts
```diff
@@ -67,6 +67,20 @@ const locale: typeof zhCN = {
         msg: {
             'filter-header-forbidden': 'You can\'t move the header row of a filter.',
         },
+        date: {
```
wzhudev, line 70: In the future we need to support countries that do not use Gregorian calendar.

packages/sheets-filter-ui/src/worker/generate-filter-values.service.ts
```diff
@@ -137,3 +133,164 @@ export function getFilterByValueItems(
 
     return items;
 }
+
+export function getFilterTreeByValueItems(
```
wzhudev, line 137: This should have test covered. For example, different number formats of cells.

wzhudev (changes requested):

packages/sheets-filter-ui/src/worker/generate-filter-values.service.ts
ybzky: different types of formats are added to the filter test demo data

wzhudev (approved):

PR #3883, opened 2024-10-29
fix: searcharray

lumixraku:
close https://github.com/dream-num/univer-pro/issues/2810

Modified the searchArray implementation to only return indices within the array. It will not return -1 or infinity.

searchArray implemention changed

## Pull Request Checklist

- [ ] multiple viewports (after sheets freeze)
- [ ] click cell after scaling and scrolling.
- [ ] select range after scaling and scrolling.
- [ ] select row & col after scaling and scrolling.

### DIFF
--- a/packages/core/src/shared/array-search.ts
+++ b/packages/core/src/shared/array-search.ts
@@ -14,6 +14,61 @@
  * limitations under the License.
  */
 
+/**
+ * Return the index of the first value in an ascending array that is greater than the target value. If there is no value greater than the target, return -1.
+ *
+ * Alternatively, you can consider inserting a number to ensure the array remains sorted, and return the position for insertion. If the target is the same as the maximum value, return -1.  (or arr.length -1)
+ * @param arr
+ * @param target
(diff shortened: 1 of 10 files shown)


DR-Univer (approved):

packages/core/src/shared/array-search.ts
```diff
@@ -14,6 +14,61 @@
...
  */
 
+/**
+ * Return the index of the first value in an ascending array that is greater than the target value. If there is no value greater than the target, return -1.
+ *
+ * Alternatively, you can consider inserting a number to ensure the array remains sorted, and return the position for insertion. If the target is the same as the maximum value, return arr.length -1
+ * @param arr
+ * @param target
+ */
+export function orderSearchArray(arr: number[], target: number) {
```
wzhudev, line 24:
```suggestion
export function searchInOrderredArray(arr: number[], target: number) {
```

wzhudev (approved):
LGTM

PR #3945, opened 2024-11-04
refactor(permission-ui): refactor permission-ui & user part can be customized

ybzky:
close #xxx

## Pull Request Checklist


Gggpound (approved):

packages/core/src/services/command/command.service.ts
```diff
@@ -373,12 +373,11 @@ export class CommandService extends Disposable implements ICommandService {
 
             throw new Error(`[CommandService]: command "${id}" is not registered.`);
         } catch (error) {
-            this._logService.error(error);
-
             if (error instanceof CustomCommandExecutionError) {
                 // If need custom logic, can add it here
                 return false as R;
             } else {
+                this._logService.error(error);
```
wzhudev, line 380: Wouldn't the error get logged twice here?

packages/sheets-ui/src/commands/commands/worksheet-protection.command.ts
```diff
@@ -37,6 +36,7 @@ export interface IDeleteWorksheetProtectionParams {
 export interface ISetWorksheetProtectionParams {
     permissionId: string;
     rule: IWorksheetProtectionRule;
+    oldRule: IWorksheetProtectionRule;
```
wzhudev, line 39: Should have some JSDoc on command params.

wzhudev (changes requested):
A few questions:

* How can users custom the users' list panel? If there a tutorial or something?
* There are two different commands with the same name `AddRangeProtectionCommand` and why?
* Can Node.js SDK users change permissions?

wzhudev:
And I have to say that the demo on preview site looks really like a bug after you set some permissions and they do not appears in the sidebar.

ybzky:
1. node supports setting permissions but has not started to be modified
2. the legacy code has been deleted
3. currently all changes to the panel are packaged in the univer-pro for QA verification because they involve request-related logic

wzhudev (approved):

PR #3967, opened 2024-11-06
fix(range-selector): range-selector can not blur

Gggpound:
close #xxx

## Pull Request Checklist

### DIFF
--- a/packages/sheets-ui/src/views/permission/panel-detail/index.tsx
+++ b/packages/sheets-ui/src/views/permission/panel-detail/index.tsx
@@ -22,7 +22,7 @@ import { deserializeRangeWithSheet, serializeRange } from '@univerjs/engine-form
 
 import { ObjectScope, UnitAction, UnitObject, UnitRole } from '@univerjs/protocol';
 import { RangeProtectionRuleModel, setEndForRange, SheetsSelectionsService, WorksheetProtectionRuleModel } from '@univerjs/sheets';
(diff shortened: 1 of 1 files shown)


packages/sheets-ui/src/views/permission/panel-detail/index.tsx
```diff
@@ -329,15 +329,16 @@ export const SheetPermissionPanelDetail = ({ fromSheetBar }: { fromSheetBar: boo
...
     };
 
-    const handlePanelClick = (e: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+    const handlePanelClick = (e: MouseEvent) => {
         const handleOutClick = rangeSelectorActionsRef.current?.handleOutClick;
-        handleOutClick && handleOutClick(e, isFocusRangeSelectorSet);
+        handleOutClick && handleOutClick(e, () => isFocusRangeSelectorSet(false));
     };
 
+    useSidebarClick(handlePanelClick);
```
wzhudev, line 337: This hook is suspicious. What about other focusable elements not in the panel, especially considering we are an SDK?

wzhudev (changes requested):
Bug confirmed. We need a more generic solution.

packages/sheets-ui/src/views/permission/panel-detail/index.tsx
Gggpound: This focused state is left entirely to the external business of the component (different cases have different defocusing moments)

PR #4043, opened 2024-11-12
fix(uni): add set-editor-resize-operation

Gggpound:
close #xxx

## Pull Request Checklist


packages-experimental/uniui/src/commands/operations/set-editor-resize.operation.ts
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+import type { IOperation } from '@univerjs/core';
+import { CommandType } from '@univerjs/core';
+
+export const SetEditorResizeOperation: IOperation = {
+    id: 'sheet.operation.set-editor-resize',
+    type: CommandType.OPERATION,
+    handler: (accessor, params) => true,
```
wzhudev, line 23: Do not use operation as events which is an anti-pattern in our codebase. This will cause performance issue since there are lots of command listeners.

wzhudev (changes requested):
Please check if there is a more direct way to do this.

packages-experimental/uniui/src/commands/operations/set-editor-resize.operation.ts
Gggpound: The code has been removed

wzhudev (approved):

PR #4174, opened 2024-11-28
refactor: rename PLUGIN_CONFIG_KEY

hexf00:
I think there may be a better way

```

export abstract class Plugin extends Disposable {
    + static configKey: string;
}


// use XXXPlugin.configKey

```

close #xxx

## Pull Request Checklist

### DIFF
--- a/packages-experimental/action-recorder/src/controllers/config.schema.ts
+++ b/packages-experimental/action-recorder/src/controllers/config.schema.ts
@@ -16,9 +16,9 @@
 
 import type { MenuConfig } from '@univerjs/ui';
 
-export const PLUGIN_CONFIG_KEY = 'action-recorder.config';
+export const ACTION_RECORDER_PLUGIN_CONFIG_KEY = 'action-recorder.config';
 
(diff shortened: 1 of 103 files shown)


jikkai:
LGTM, but all the remaining packages need to be modified.

hexf00:
> LGTM, but all the remaining packages need to be modified.

If you are referring to the `PLUGIN_CONFIG_KEY` of other plugins, I don't think we should introduce a new export in the refactoring as other `PLUGIN_CONFIG_KEY` have not been exported yet

jikkai:
> > LGTM, but all the remaining packages need to be modified.
> 
> If you are referring to the `PLUGIN_CONFIG_KEY` of other plugins, I don't think we should introduce a new export in the refactoring as other `PLUGIN_CONFIG_KEY` have not been exported yet

What I mean is, directly rename all the `PLUGIN_CONFIG_KEY` to avoid duplicate name issues when exporting in the future.

jikkai (approved):

PR #4183, opened 2024-11-28
refactor(drawing): use core exports instead of drawing exports

hexf00:
close #xxx

## Pull Request Checklist


packages/drawing/src/services/interface.ts
```diff
@@ -14,14 +14,4 @@
...
-export interface IRotationSkewFlipTransform {
-    angle?: number;
-    skewX?: number;
-    skewY?: number;
-    flipX?: boolean;
-    flipY?: boolean;
-}
-
-export interface ITransformState extends IAbsoluteTransform, IRotationSkewFlipTransform {}
+export type { IRotationSkewFlipTransform, ITransformState } from '@univerjs/core';
```
jikkai, line 17: I don't think there's a need to export the stuff from the core package here anymore.

jikkai (changes requested):
If not exporting these duplicate variables causes issues, it indicates that the code importing them is flawed and should be corrected.

packages/drawing/src/services/interface.ts
hexf00: Use Core exports instead of drawing exports.

packages/docs-drawing-ui/src/commands/commands/set-drawing-arrange.command.ts
```diff
@@ -15,17 +15,16 @@
...
+import type { IDrawingJsonUndo1, IDrawingOrderMapParam } from '@univerjs/drawing';
 import {
+    ArrangeTypeEnum,
     CommandType,
     ICommandService,
     JSONX,
-    Tools,
-} from '@univerjs/core';
-import type { IRichTextEditingMutationParams } from '@univerjs/docs';
+    Tools } from '@univerjs/core';
```
jikkai, line 25: A line break missing here

jikkai (approved):

PR #4280, opened 2024-12-11
Update ToolbarButton.tsx

strawberry2020:
The default value of button type in the browser is submit. If you embed univer in form, the click event of the ToolbarButton will trigger the submit event of the form. type should be specified as botton by default.


wzhudev:
@jikkai Please take a look into this as this has been stale for a while.

PR #4302, opened 2024-12-14
feat(textsplit): support text split util function

VicKun4937:
# done
- support text split  to column
- auto find the delimiter
- custom delimiter
- treat multiple delimiters as one
# todo
- auto insert column
- cell format
- mutation and command wrapper


packages/sheets/src/controllers/basic-worksheet.controller.ts
```diff
@@ -300,6 +301,9 @@ export class BasicWorksheetController extends Disposable implements IDisposable
                 ToggleCellCheckboxCommand,
                 SetWorksheetDefaultStyleMutation,
                 SetWorksheetDefaultStyleCommand,
+
+                // split text to columns
```
wzhudev, line 305: This comment can be ommited.

wzhudev (approved):
LGTM!

PR #4372, opened 2024-12-26
fix:  better disable selection ---> transparent selection

lumixraku:
close #xxx

Add two more API, transparentSelection & showSelection.
May be more easily to use.

Selection function still works, but you can not see them. 
Users can still edit cells,  select diffrent cells. 

Seleciton functions is not working after previous API "disableSelection"
That means users can not even select different cells. 

## Pull Request Checklist


packages/sheets-formula-ui/src/services/render-services/ref-selections.render-service.ts
```diff
@@ -44,13 +44,15 @@ export class RefSelectionsRenderService extends BaseSelectionRenderService imple
         @Inject(ThemeService) themeService: ThemeService,
         @IShortcutService shortcutService: IShortcutService,
         @Inject(SheetSkeletonManagerService) sheetSkeletonManagerService: SheetSkeletonManagerService,
+        @IContextService private readonly _contextService: IContextService,
```
wzhudev, line 47: It should be a `protected` value in the parent class please you pass it to the `super`.

wzhudev (approved):
Overall LGTM.

weird94 (approved):

hexf00 (approved):

PR #4438, opened 2025-01-06
fix(formula): fix formula facade import

hexf00:
close #xxx

## Pull Request Checklist


examples/esbuild.config.ts
```diff
@@ -41,6 +44,104 @@ const monacoEditorEntryPoints = [
     'vs/editor/editor.worker.js',
 ];
 
+/**
+ * fix `import '@univerjs/xxx/lib/index.css'` not work on source code dev mode
+ *
+ * The `stylePlugin` must be loaded after the `skipLibCssEsbuildPlugin`.
+ */
+const skipLibCssEsbuildPlugin = {
```
wzhudev, line 52: What's the context of this change?

hexf00: Added logic to directly test the packaged artifacts in the univer repository.
`lib/index.css` is not available in the development environment and needs to be skipped; otherwise, it will cause errors.

wzhudev: I see. Please add documentation in the repository of how to use this. A new package.json scripts would also be good. You can use [cross-env](https://www.npmjs.com/package/cross-env) to simplify things.

jikkai (approved):

PR #4540, opened 2025-01-21
fix(sheet): support read default style when create editor

VicKun4937:
- in editor , the default style should be read

### DIFF
--- a/packages/sheets-ui/src/services/editor-bridge.service.ts
+++ b/packages/sheets-ui/src/services/editor-bridge.service.ts
@@ -355,8 +355,9 @@ export class EditorBridgeService extends Disposable implements IEditorBridgeServ
             worksheet.getCell(startRow, startColumn),
             location
         );
+        const cellDefaultStyle = worksheet.getCellDefaultStyle(startRow, startColumn, skeleton.isRowStylePrecedeColumnStyle);
 
-        documentLayoutObject = cell && skeleton.getCellDocumentModelWithFormula(cell);
+        documentLayoutObject = cell && skeleton.getCellDocumentModelWithFormula(cell, cellDefaultStyle);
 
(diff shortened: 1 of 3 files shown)


packages/engine-render/src/components/sheets/sheet-skeleton.ts
```diff
@@ -1532,7 +1538,7 @@ export class SpreadsheetSkeleton extends Skeleton {
             ...options,
         };
 
-        const style = this._styles.getStyleByCell(cell);
+        const style = { ...options.cellDefaultStyle, ...this._styles.getStyleByCell(cell) };
```
wzhudev, line 1541: When should we use `composeStyles`?

packages/sheets-ui/src/services/editor-bridge.service.ts
```diff
@@ -340,8 +340,9 @@ export class EditorBridgeService extends Disposable implements IEditorBridgeServ
             worksheet.getCell(startRow, startColumn),
             location
         );
+        const cellDefaultStyle = worksheet.getCellDefaultStyle(startRow, startColumn, skeleton.isRowStylePrecedeColumnStyle);
 
-        documentLayoutObject = cell && skeleton.getCellDocumentModelWithFormula(cell);
+        documentLayoutObject = cell && skeleton.getCellDocumentModelWithFormula(cell, cellDefaultStyle);
```
wzhudev, line 345: Is this the only place that `cellDefaultStyle` should be passed? `getCellDocumentModelWithFormula` are called in other situations such as removing cell hyperlink.

VicKun4937:
this pr seems not necessary one, considering close it.

wzhudev:
> this pr seems not necessary one, considering close it.

Have you made a conclusion here? Shall we close it?

wzhudev:
Resolved in another PR.

PR #4573, opened 2025-01-25
fix(statusbar): fix status bar can not read date type cell

VicKun4937:
close https://github.com/dream-num/univer-pro/issues/3073


wzhudev (approved):
LGTM

PR #4591, opened 2025-02-07
fix: fix multi memory leak problems

wzhudev:
close #xxx

## Pull Request Checklist

### DIFF
--- a/packages/sheets-ui/src/controllers/clipboard/clipboard.controller.ts
+++ b/packages/sheets-ui/src/controllers/clipboard/clipboard.controller.ts
@@ -148,7 +148,7 @@ export class SheetClipboardController extends RxDisposable {
         super();
         this._init();
         this._initCommandListener();
-        // this._initUIComponents();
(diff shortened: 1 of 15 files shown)


packages/sheets-ui/src/controllers/clipboard/clipboard.controller.ts
```diff
@@ -148,7 +148,7 @@ export class SheetClipboardController extends RxDisposable {
         super();
         this._init();
         this._initCommandListener();
-        this._initUIComponents();
+        // this._initUIComponents();
```
lumixraku, line 151: no need ui comp?

Dushusir (approved):

packages/sheets-ui/src/controllers/clipboard/clipboard.controller.ts
wzhudev: Good catch.

PR #4658, opened 2025-02-17
feat(formula): add custom data to value objects

jcalem:
close #xxx

This PR adds custom data as an attribute to `BaseValueObject` class to enable metadata to be included in custom function implementations

Usage:
```
export class CustomFunction extends BaseFunction {
  override calculate(value: BaseValueObject) {
     return StringValueObject.create('TEST').withCustomData({key: 'value'});
  }
}
```

## Pull Request Checklist


wzhudev:
Hi. Thanks for opening this PR. Could you please give us an example how this feature will be used? Thanks.

jcalem:
> Hi. Thanks for opening this PR. Could you please give us an example how this feature will be used? Thanks.

Absolutely. As mentioned in the description, this PR enables metadata to be returned in custom function implementations. So for example, let's say I'm creating a custom function that calls a backend API to retrieve the weather forecast in a city. With this change, I can attach metadata like so:

```
export class GetWeatherFunction extends BaseFunction {
  override calculate(cityObject: BaseValueObject) {
     return new AsyncObject(invokeAPI(cityObject));
  }
}

async function invokeAPI(cityObject: BaseValueObject) {
  try {
   const city = cityObject.getValue();
    const response = await ApiService.get(`/get/${city}`);
    const metadata = {
     low: response.low,
     high: response.high,
     source: response.source
   }
    return StringValueObject.create(response.current).withCustomData(metadata);
  } catch (error) {
    return new ErrorValueObject(ErrorType.VALUE);
  }
}
```

So invoking this custom function would set the cell value to `response.current` (the current weather in the city), but also attach the low, high, and source as metadata to the cell. The metadata can be used to display information is other components, such as a custom side panel.

Let me know if you have any other questions!

packages/engine-formula/src/engine/value-object/base-value-object.ts
```diff
@@ -33,6 +33,8 @@ export interface IArrayValueObject {
     column: number;
 }
 export class BaseValueObject extends ObjectClassType {
+    private _customData: any;
```
Dushusir, line 36: Avoid using any if possible

packages/engine-formula/src/engine/value-object/base-value-object.ts
```diff
@@ -72,6 +74,15 @@ export class BaseValueObject extends ObjectClassType {
         /** abstract */
     }
 
+    withCustomData(data: any) {
```
Dushusir, line 77: Avoid using any if possible

jcalem: Updated to `CustomData` type

packages/engine-formula/src/engine/value-object/base-value-object.ts
jcalem: Updated to `CustomData` type

Dushusir (approved):

PR #4771, opened 2025-03-04
fix(sheets-conditional-formatting): correct bottom percent rank calculation

maxsbelt:
Fixed incorrect highlight for conditional formatting in case of rank bottom percent. On example highlight should be applied only for one cell (10% of 10 items). Before the change there were two cells highlighted, after - only one. Also added test that checks that 2 cells of 5 are highlighted for rank bottom 40%.

|before|after|
|-|-|
|Screenshot 2025-03-04 at 18 30 34|Screenshot 2025-03-04 at 18 32 11|

## Pull Request Checklist


wzhudev:
Thanks for opening this PR. We really appreciate your contribution to Univer.

image

There is another case which should be take into consideration:

image

In this case, `1` should be highlighted. Please refer to Excel as a reference.

wzhudev:
Thanks for opening this PR. We really appreciate your effort!

There is another corner case maybe you should take into consideration. In this case, 0.9 should be highlighted

image

Compared to Excel:

image

packages/sheets-conditional-formatting/src/models/calculate-unit-v2/highlight-cell-calculate-unit.ts
```diff
@@ -84,7 +84,7 @@ export class HighlightCellCalculateUnit extends BaseCalculateUnit<Nullable<IConf
                     allValue.sort((a, b) => b - a);
                     const configRule = context.rule.rule as IRankHighlightCell;
                     const targetIndex = configRule.isPercent
-                        ? Math.floor(Math.max(Math.min(configRule.value, 100), 0) / 100 * allValue.length)
```
wzhudev, line 87: You can use `Tools.clamp` to replace `Math.max(Math.min())`.

maxsbelt:
> There is another corner case maybe you should take into consideration. In this case, 1 should be highlighted

Good catch. Actually after multiple tests I end up with logic that more complicated than initial fix. But not should work properly. I have added test for corner case.

wzhudev:
Hi. Your original solution is quiet good. Just need to make sure that the value is not out of bounds. I've made some updates based on your first solution and added a test.

Really appreciate your effort!

wzhudev (approved):

PR #5197, opened 2025-05-14
fix(color): fix short hex color error & add unit test

VicKun4937:
- fix color render error abot "#ccc"

### DIFF
--- a/packages/engine-render/src/services/canvas-color.service.ts
+++ b/packages/engine-render/src/services/canvas-color.service.ts
@@ -63,13 +63,13 @@ export class CanvasColorService extends Disposable implements ICanvasColorServic
             cachedColor = rgbToHex(invertedColor);
 
             // append alpha channel if the original has alpha channel
-            if (color.length === 4) {
-                const alpha = color.charAt(3);
-                cachedColor += alpha + alpha;
-            }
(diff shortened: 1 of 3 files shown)


packages/engine-render/src/services/canvas-color.service.ts
```diff
@@ -63,13 +63,13 @@ export class CanvasColorService extends Disposable implements ICanvasColorServic
             cachedColor = rgbToHex(invertedColor);
 
             // append alpha channel if the original has alpha channel
-            if (color.length === 4) {
```
wzhudev, line 66: `#RGBA` 也是一种合法的格式，可以考虑兼容

wzhudev (approved):
LGTM

packages/engine-render/src/services/canvas-color.service.ts
VicKun4937: if a hex color with alpha , it can not be a short one, it must be long one. such we can use `#ccccccff`  but not `#cccf`

VicKun4937: supported.  It should be length 5, becuase it startwith `#`

PR #5339, opened 2025-06-07
fix(conditional-formatting): merge CF styles with interceptorStyle to override number format colors

### LINKED ISSUE #5338: [Bug] Conditional formatting text color not applied to negative values with custom number formats

0.8.0

### Reproduction link

To reproduce : open a new sheet, write a negative value into a cell, change the number format to 'Number' instead of 'General', then apply a conditional formatting 'number less than 0' with a colored text and the color will not be applied.

I was able to reproduce with the basic demo here : https://docs.univer.ai/en-US/playground/sheets/basic

Image

### Expected behavior

The text should be colored according to the condition.

### Actual behavior

The text is not colored. Only the filling color is applied.

### System information

_No response_

bjaq:
When a cell has both conditional formatting and a custom number format, the number format's color was overriding the conditional formatting text color because interceptorStyle takes precedence over regular styles during rendering.

This fix ensures conditional formatting styles are merged into interceptorStyle when needed, allowing CF text colors to be applied correctly to negative values with custom number formats.

close #5338

## Pull Request Checklist

- [X] Related tickets or issues have been linked in the PR description (or missing issue).
- [X] [Naming convention](https://github.com/dream-num/univer/blob/dev/docs/NAMING_CONVENTION.md) is followed (**do please** check it especially when you created new plugins, commands and resources).
- [X] Unit tests have been added for the changes (if applicable).
- [X] Breaking changes have been documented (or no breaking changes introduced in this PR).

### DIFF
--- a/packages/sheets-conditional-formatting-ui/src/controllers/cf.render.controller.ts
+++ b/packages/sheets-conditional-formatting-ui/src/controllers/cf.render.controller.ts
@@ -80,6 +80,14 @@ export class SheetsCfRenderController extends Disposable {
                         ...result.style,
                     };
                     Object.assign(cloneCell, { s: activeStyle });
+                    
+                    // Merge conditional formatting styles with existing interceptorStyle to ensure CF text color takes precedence
+                    if (cloneCell.interceptorStyle || result.style.cl) {
+                        cloneCell.interceptorStyle = {
+                            ...cloneCell.interceptorStyle,
+                            ...result.style,
+                        };
+                    }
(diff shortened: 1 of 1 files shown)


jikkai:
Thanks for your PR ❤️
Please confirm whether the issue has been resolved after checking this reply: https://github.com/dream-num/univer/issues/5338#issuecomment-2954635009.

bjaq:
I'm closing since it has already been solved by #5344

PR #5601, opened 2025-07-28
fix(table): table name should not include space

VicKun4937:
- in excel , table name can be used in formula , so we should keeep the name not with blank or space


packages/sheets-table/src/commands/commands/add-sheet-table.command.ts
```diff
@@ -30,6 +30,10 @@ export interface IAddSheetTableCommandParams {
     options?: ITableOptions;
 }
 
+function checkSpaceInTableName(name: string): boolean {
```
jikkai, line 33: Would using regular expressions be a better choice?
Abstracting this check into a standalone function adds no real value.

jikkai (changes requested):
Should not throw an error within a command

packages/sheets-table/src/commands/commands/add-sheet-table.command.ts
VicKun4937: I have moved it to core package, it can be use for custom name both.

jikkai (approved):

PR #5608, opened 2025-07-29
feat: locale compatibility & date formula; fix duplicate `'required'`

kenny-not-dead:
add ru locale compatibility & date formula, fix duplicate 'required' in mounth formula


jikkai (approved):
LGTM

PR #5764, opened 2025-08-28
fix: insert cols&rows from rigth mouse btn menu

### LINKED ISSUE #5762: [Bug] Insert col&row

0.10.5

### Reproduction link

when using the menu via the right mouse button, if you insert columns or rows, the selected number does not depend on the result, only one is always inserted

https://github.com/user-attachments/assets/b9312b06-21e3-461c-9da6-505a9a1a2347

### Expected behavior

after selecting insert row or column, the previously selected number of columns or rows appears

### Actual behavior

after selecting insert row or column only one column or row appears

### System information

_No response_

kenny-not-dead:
close #5762 

Considering that ui provides the ability to select the number of inserted columns and rows, I fixed this option

right-click on a cell, select the desired number of rows or columns, click on this button

I attached the result of the changes to the video, it works as expected
[Screencast from 2025-08-28 21-34-00.webm](https://github.com/user-attachments/assets/92d8fe9d-34d5-4b52-935a-683064a7c5ce)

### DIFF
--- a/packages/sheets/src/commands/commands/insert-row-col.command.ts
+++ b/packages/sheets/src/commands/commands/insert-row-col.command.ts
@@ -175,7 +175,7 @@ export const InsertRowByRangeCommand: ICommand = {
 export const InsertRowBeforeCommand: ICommand = {
     type: CommandType.COMMAND,
     id: 'sheet.command.insert-row-before',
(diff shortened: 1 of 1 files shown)


kenny-not-dead:
corrected the tests

PR #5793, opened 2025-09-03
feat: double line

kenny-not-dead:
in continuation of the underlining, i decided to finish the question with a double line. Also, i made the part i added yesterday in the demo documents a bit cleaner, i think it was too bulky even for a temporary demonstration.

in fact, I have a question about TextDecoration.NONE. Can you clarify what was meant by this concept, if it is simply its absence, then perhaps it may look strange, show: true, type: none? or was it a trigger in the command that would set show false? Also I didn't quite understand how TextDecoration.WORDS can be displayed.  As far as I could see in this repository, these values ​​are not used.

start demo => 
1. open sheet
2. open doc

After: -->
image
image

### DIFF
--- a/mockdata/src/docs/default-document.data-simple.ts
+++ b/mockdata/src/docs/default-document.data-simple.ts
@@ -54,26 +54,11 @@ const exampleTables = [
 
 const title = 'Examples of Accessible Data Tables\r';
 const description = 'Basic Data Table with Column Headings\r';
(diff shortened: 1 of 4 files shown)


kenny-not-dead:
@jikkai 
Please clarify if everything is okay with this PR or maybe something needs to be corrected on my part?

PR #5818, opened 2025-09-07
feat: underline selection

kenny-not-dead:
close #xxx

I continued implementing text underlining, considering the possibility of installing different options for underlining, it seemed logical to me to have a drop-down list of possible options

If this implementation is suitable, I can add this in sheets

I hope you like this implementation, I personally like it =)

open demo  -> try use underline

[Screencast from 2025-09-07 21-07-33.webm](https://github.com/user-attachments/assets/a4884148-c71e-42f8-8ec1-9d30d18c3e2b)

## Pull Request Checklist

### DIFF
--- a/packages/docs-ui/src/commands/commands/__tests__/inline-format.command.spec.ts
+++ b/packages/docs-ui/src/commands/commands/__tests__/inline-format.command.spec.ts
@@ -15,7 +15,15 @@
  */
 
 import type { DocumentDataModel, ICommand, Injector, IStyleBase, Univer } from '@univerjs/core';
-import { BooleanNumber, ICommandService, IUniverInstanceService, RedoCommand, UndoCommand, UniverInstanceType } from '@univerjs/core';
+import {
+    BooleanNumber,
+    ICommandService,
+    IUniverInstanceService,
+    RedoCommand,
+    TextDecoration,
+    UndoCommand,
+    UniverInstanceType,
+} from '@univerjs/core';
 import { DocSelectionManagerService, RichTextEditingMutation, SetTextSelectionsOperation } from '@univerjs/docs';
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import {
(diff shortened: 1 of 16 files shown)


kenny-not-dead:
@jikkai 
please clarify can i leave a TODO or should i delete it? i will probably add it myself, but a bit later. i would also like to clarify about the message from the bot about test coverage, i didn't see any tests on the menu here, similar to sheets

jikkai:
Apologies for the late reply. I’ve been handling a few other priorities 🤧
Thank you, as always, for your continued support 🫶

From a review perspective your implementation looks perfectly fine. However, as it may relate to Univer Docs import/export, we’ll need a little extra time to review the implications :)

kenny-not-dead:
@jikkai  I understand you, I will wait for an answer on this issue =)

Considering that I also did import and export on my backend, I think it will not affect you for long, since your underscore types were initially well designed. I think the only issue is in supporting the necessary switch-case types

kenny-not-dead:
@jikkai 

I decided to expand on this PR. First, I decided to make it more similar to other dropdowns, as I hadn't previously realized that the hover value applies to the entire element. I also expanded dark theme support and added two type variants. Overall, I decided to make this component more similar to a table component.

image

image

kenny-not-dead:
It's weird, this test used to work fine, but I don't think I added anything that would have broken it.

jikkai:
> It's weird, this test used to work fine, but I don't think I added anything that would have broken it.

This is an issue caused by pixel comparison. As long as the content in the examples of docs or sheets is modified, resulting in changes to the rendering of the visible area, this problem will be triggered.
This usually does not constitute a real issue, we will fix it after merging :)

kenny-not-dead:
@jikkai 

Thanks, then I'll wait for confirmation of this decision. I think it's gotten better after the last change. I also have a suggestion regarding underscores in tables, since Excel also supports double underscores.

kenny-not-dead:
@VicKun4937 
Please clarify if this PR is still relevant?

kenny-not-dead:
I think this can be closed...

PR #5834, opened 2025-09-10
fix(docs): active bg color

kenny-not-dead:
in continuation of previous PR added active background color for documents

[Screencast from 2025-09-10 21-14-11.webm](https://github.com/user-attachments/assets/2f230baa-5279-44e5-b055-488113a55662)


kenny-not-dead:
@jikkai 
Please clarify if this PR has any problem on my side?

PR #5841, opened 2025-09-11
chore: change depricated getUniverDocInstance to getUnit

kenny-not-dead:
I just decided to remove the deprecated getUniverDocInstance, now it is not used in the code, except for UniverInstanceService


kenny-not-dead:
I don't quite understand why the tests fail, it seems like the same getUnit is actually called 
getUniverDocInstance(unitId: string): Nullable<DocumentDataModel> {
   return this.getUnit<DocumentDataModel>(unitId, UniverInstanceType.UNIVER_DOC);
}

PR #5855, opened 2025-09-13
fix: active cell border color

kenny-not-dead:
the color in the color picker for cells that have borders has always been default. Now it has the same behavior as other color pickers

Before: always default color #000000

After: 

[Screencast from 2025-09-13 21-25-20.webm](https://github.com/user-attachments/assets/614d2961-2ed0-46bd-b9ac-43af6ca0a036)

## Pull Request Checklist


kenny-not-dead:
@jikkai 
sorry, please clarify if this is some problem on my part? Also I was planning to offer highlighting of active type of border highlighting if you support it?

PR #5879, opened 2025-09-17
feat: diagonal double border

### LINKED ISSUE #5878: [Feature] double  line  enhance
### Initial checklist

### Problem

In excel , the double line draws outside the grid line, here is some test:

Image

we can find : 
1. on the top of double line , only inside part be drawed
2. the right part cell border will overwrite the left one

kenny-not-dead:
close #5878

This is not a full implementation of the features that were proposed @VicKun4937, because it seems to me that this requires some decomposition into subtasks. However, I didn't know if I should include the issue number in the close. I decided to point it out just in case. When closing, it might be worth resuming the task of double underlines not related to diagonals.

But I would like to note that I have just started to try out the implementation related to this feature, and I still can’t imagine how I can remove the cross highlight between the double lines.  Perhaps you have an idea on this matter. 
I'll be glad to hear it if that's the case.

If we are closer to this PR, then it seemed necessary to me to first implement the functionality for double borders associated with diagonals

Pasted image (2)

Pasted image (3)

Pasted image

## Pull Request Checklist

### DIFF
--- a/packages/engine-render/src/basics/draw.ts
+++ b/packages/engine-render/src/basics/draw.ts
@@ -97,7 +97,7 @@ export function drawLineByBorderType(ctx: UniverRenderingContext, type: BORDER_L
     ctx.closePathByEnv();
 }
 
-export function drawDiagonalLineByBorderType(ctx: UniverRenderingContext, type: BORDER_LTRB, position: IPosition) {
+function drawDiagonalSingleLineByBorderType(ctx: UniverRenderingContext, type: BORDER_LTRB, position: IPosition) {
     let drawStartX = 0;
     let drawStartY = 0;
     let drawEndX = 0;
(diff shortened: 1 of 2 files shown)


VicKun4937:
@kenny-not-dead  About " how to remove the cross highlight between the double lines", maybe you can start draw the double line by four part triangles, maybe the following steps can work:
47ea9c6e-3499-4b51-a835-d045a8f189c8

calculate the diagonal slope ， then get the offset, in this way , we can get the the piont P1
here I should explain how to calc the offset by Trigonometric functions

4026cdb4-2737-493a-ba15-afc2d9360947
the slope angle of a straight line  we can easy get by function `a = tan(α)` and `α = arctan(a)`, then we know `sin(α) = offset/ h`

by above , we can get the offset , it means piont P1 (x, y) = A(x, y + offset), if the P1,P2 you got the cross point we can get also 

At last , we just use canvs `moveTo`, `lineTo`.

**All above it complex , I think is unnecessary, just show a way to solve your problem. I think cross in center is acceptable**

VicKun4937:
> @kenny-not-dead About " how to remove the cross highlight between the double lines", maybe you can start draw the double line by four part triangles, maybe the following steps can work: 

image

 calculate the diagonal slope ， then get the offset, in this way , we can get the the piont P1 here I should explain how to calc the offset by Trigonometric functions
image

image


 by above , we can get the offset , it means piont P1 (x, y) = A(x, y + offset), if the P1,P2 you got the cross point we can get also
> 
> At last , we just use canvs `moveTo`, `lineTo`.
> 
> **All above it complex , I think is unnecessary, just show a way to solve your problem. I think cross in center is acceptable**

kenny-not-dead:
@VicKun4937 

Sorry, I probably expressed myself incorrectly about the cross-highlighting. I meant the table skeleton that's between the two lines. This isn't a problem with diagonal lines. That is, if I move the outer border beyond the cell, that line becomes visible.

image

image

VicKun4937:
> > @kenny-not-dead About " how to remove the cross highlight between the double lines", maybe you can start draw the double line by four part triangles, maybe the following steps can work:
> 
> image
> calculate the diagonal slope ， then get the offset, in this way , we can get the the piont P1 here I should explain how to calc the offset by Trigonometric functions image
> 
> image
> by above , we can get the offset , it means piont P1 (x, y) = A(x, y + offset), if the P1,P2 you got the cross point we can get also
> 
> > At last , we just use canvs `moveTo`, `lineTo`.
> > **All above it complex , I think is unnecessary, just show a way to solve your problem. I think cross in center is acceptable**



> @VicKun4937
> 
> Sorry, I probably expressed myself incorrectly about the cross-highlighting. I meant the table skeleton that's between the two lines. This isn't a problem with diagonal lines. That is, if I move the outer border beyond the cell, that line becomes visible.
> 
> image image

sorry for misunderstood your question.
 Because in univer  when a cell is drawed, the `ctx.clip();` will be used for drawing the cell,  but when text overflow it will be not clip, it is complex I think , you can update your code to fix the CI error, and maybe I will finish it in the future

kenny-not-dead:
@VicKun4937 
I'll update this soon.  thanks for your advice

kenny-not-dead:
@VicKun4937 

I fixed it, sorry for the delay

VicKun4937 (approved):

PR #5891, opened 2025-09-19
fix: shortcut cropped content in kbd component

kenny-not-dead:
close #xxx

I noticed that the key combination for long texts has a hyphenation, which makes it impossible to read the correct key combination

I fixed this issue by initially leaving the text on a single line, but it became virtually impossible to read the description of this text; it was too hidden. Therefore, I would like to suggest, in addition to fixing the key combination itself, increasing the description text to two lines. I think this will make it more convenient for users.

The problem affected the following languages: ru, fr, es, ca

I also changed the Russian language a little, because this previous translation is not actually used in conversation and is aimed at something slightly different

Before:

image

image

image

image

image

After: 

image

image

image

image

image

image

## Pull Request Checklist

### DIFF
--- a/packages/design/src/components/kbd/Kbd.tsx
+++ b/packages/design/src/components/kbd/Kbd.tsx
@@ -31,8 +31,8 @@ export function KBD(props: IKBDProps) {
     return (
         <span
             className={clsx(`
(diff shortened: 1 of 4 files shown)


jikkai (approved):
LGTM

PR #5906, opened 2025-09-22
feat: computed rotate angle

kenny-not-dead:
close #xxx

It might be worth considering calculating the rotation angle, since currently it's possible to enter any number from -3600 to 3600, but does that make sense? I suggested resetting the angle to 0 when the value is 360 or -360, and also calculating the rotation angle by subtracting excess revolutions.

[Screencast from 2025-09-22 19-28-45.webm](https://github.com/user-attachments/assets/ae81e48f-f3e8-4d16-aa25-76207c3ed95d)

## Pull Request Checklist


wpxp123456:
Thanks.
But we believe there is no need to convert the values ​​that exceed the limit, just limit the range of values.
Do you think this PR is okay?
https://github.com/dream-num/univer/pull/5914

kenny-not-dead:
@wpxp123456 Yes, I like the idea of ​​a minimum and maximum value, like yours, even more. Indeed, I thought about it when creating this PR, but given that it was previously in the -3600 - 3600 range, I was hesitant to suggest limiting it via a min/max. I'll close the current PR, then, since yours is the most suitable. Thanks.

wpxp123456:
Okay, thank you very much.

PR #5973, opened 2025-10-14
fix(sheet): fix some memory leak error

VicKun4937:
- @jikkai  here is last one memory leak error
- maybe we should release 
企业微信截图_17604504518924


VicKun4937:
if i replace the function deriveStateFromActiveSheet$  to return of(defaultValue),  the workbook will be release

image

PR #6002, opened 2025-10-20
fix: err with updateInlineDrawingCoordsAndBorder

kenny-not-dead:
close #xxx

With certain copy-paste combinations, an error appeared in the console
image

Before:
image
After:

no error

## Pull Request Checklist

### DIFF
--- a/packages/engine-render/src/components/docs/layout/tools.ts
+++ b/packages/engine-render/src/components/docs/layout/tools.ts
@@ -486,7 +486,7 @@ export function updateInlineDrawingCoordsAndBorder(ctx: ILayoutContext, pages: I
             const lastDivide = line.divides[line.divides.length - 1];
             const lastGlyph = lastDivide.glyphGroup[lastDivide.glyphGroup.length - 1];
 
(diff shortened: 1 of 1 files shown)


jikkai:
How can I reproduce this error?

kenny-not-dead:
@jikkai 
For your convenience, I would like to attach a video.
[Screencast from 2025-10-20 17-16-14.webm](https://github.com/user-attachments/assets/04476a7e-598b-4c7d-96f4-eecc36e43cc6)

kenny-not-dead:
@jikkai 

I did a little digging to reproduce this error in the console. I came to the conclusion that the specific buffer size doesn't affect it, but the location of the cursor after copying does... Take the text to copy (at least 2 lines), move the cursor closer to the table so that when pasting, the text ends up in the table in the demo version:

image

After copying, pay attention to where the cursor ended up and the error appeared.

image

kenny-not-dead:
@jikkai 
Please note that with this error the text is not copied, but after the fix, it is copied, i.e. this fix cures not only the error, but also the event itself

PR #6010, opened 2025-10-21
fix: delete sheet get location err

kenny-not-dead:
close #xxx

When deleting a sheet, the workbook stops displaying. I looked at the changelog and came to the conclusion that this is related to memory leak optimization https://github.com/dream-num/univer/pull/5988 At this time I decided to suggest commenting this out and giving a comment in the code about it. 
I wanted to clarify if you're only releasing this on Saturday, as it seems to me that this is a pretty disruptive change, and perhaps a patch would be appropriate. What do you think?

Before:
[Screencast from 2025-10-21 12-16-20.webm](https://github.com/user-attachments/assets/c92aebdd-39cd-422e-ab21-627f7fd5fad6)

After: 

works as before

## Pull Request Checklist


kenny-not-dead:
I see that you fixed it here, I'll close this PR then https://github.com/dream-num/univer/pull/6005

Issue #5967, opened 2025-10-13
[Bug]  Сopying a sheet and styles


kenny-not-dead:
### Before you submit this issue, have you checked the following

- [x] Is this really a problem?
- [x] I have searched the Github Issues for similar issues, but did not find anything.

### Affected packages and versions

latest

### Reproduction link

I hope I can describe this problem correctly, perhaps you may have similar behavior on your backend

When I enter data or apply a set of styles, when I subscribe to events, sheet.mutation.set-range-values ​​is generated. For example, taking applying a background, the event would look something like this: 

<img width="627" height="302" alt="Image" src="https://github.com/user-attachments/assets/b22d068c-302c-4990-88c9-98395d633ba4" />

As we can see from this, there is information only about the style itself (without any style ID), and there are no events about creating a style by ID, which I will draw attention to below.

After clicking on Copy Sheet, the sheet.mutation.insert-sheet event is generated, the event would look something like this: 

<img width="794" height="699" alt="Image" src="https://github.com/user-attachments/assets/b0e5d20d-728f-4a0d-9b8b-f7af786e3a37" />

Here it is necessary to pay attention to this part

<img width="306" height="111" alt="Image" src="https://github.com/user-attachments/assets/07d010e2-d423-459f-848d-fb72dfac5c4c" />

As you can see, the style in this cell was set by ID, by the way, it is the same cell to which we applied the background above. but, in this event there is nothing about this style by ID, there is no event before or after that could intificate it either

thus, it is currently impossible to obtain any information about the created style by its ID from events

I think this can be solved in two ways:
- adding a separate event for the created style, specifying its ID
- adding an array of styles by ID to the existing copy sheet event so it can be compared
- replacing the style ID with specific style values

I could be wrong, but it seems to me that you may have a similar problem with your backend, especially if change events are recorded by the backend based on events received via a web socket.

### Expected behavior

When reading events over a web socket, you can understand the generated style by its ID.

### Actual behavior

When copying a sheet, the style is written as an ID, however, there are no events of its own or mentions in the event

wpxp123456:
I'm sorry, I don't quite understand what you mean. Is there any difference between the style being an ID or a specific style value? If you need to determine a certain style, you can also get the specific value of the style by the ID.

kenny-not-dead:
[...]

PR #6011, opened 2025-10-21
fix: add style in copy-sheet event

kenny-not-dead:
close #xxx

In this issue https://github.com/dream-num/univer/issues/5967,  we discussed the problem of passing a style ID that collaborators might not receive in events. After which, @wpxp123456 made some changes, and overall, it's great, but unfortunately, these styles were never finished, this fix is ​​aimed at that
Sorry for my overactivity, but it seems like it's worth finishing

## Pull Request Checklist

### DIFF
--- a/packages/sheets/src/commands/commands/copy-worksheet.command.ts
+++ b/packages/sheets/src/commands/commands/copy-worksheet.command.ts
@@ -56,11 +56,13 @@ export const CopySheetCommand: ICommand = {
         config.name = getCopyUniqueSheetName(workbook, localeService, config.name);
         config.id = generateRandomId();
         const sheetIndex = workbook.getSheetIndex(worksheet);
+        const styles = workbook.getStyles().toJSON();
 
         const insertSheetMutationParams: IInsertSheetMutationParams = {
             index: sheetIndex + 1,
             sheet: config,
             unitId,
+            styles,
         };
 
         const removeSheetMutationParams: IRemoveSheetMutationParams = InsertSheetUndoMutationFactory(


wpxp123456:
Isn't this problem solved? I don't quite understand why the style of the entire workbook is placed in it.

wpxp123456:
This mutation has the worksheet's styles passed in.
image
image
Don't these meet your needs?

kenny-not-dead:
Yes, unfortunately, when I updated the version after your fix, I couldn't fix it, since in fact, the styles from the event source did not appear in the event.  until now, when an event leaves the source, it has a style as an id, but there is no information about this style with this id 
in fact, your previously made changes cannot work because styles are not attached to the copy event

kenny-not-dead:
If necessary, I can describe the flow and the problem again.

wpxp123456:
Aren't you using @univerjs-pro/collaboration @univerjs-pro/collaboration-client @univerjs-pro/collaboration-client-ui ?

kenny-not-dead:
No, I wrote my own backend, but I think my problems should be the same as yours 😊

wpxp123456:
Then let me give you an idea. Collaboration-related matters should be handled by the collaboration side. The SDK has reserved the styles field.

You can write a mutation interceptor and insert the copied worksheet style before sending it to the socket. You can refer to the interceptor related code in OSS.

I'm Sorry, the pro code cannot be leaked, it is the company's property.

wpxp123456:
Mainly `univerInstanceService`, through which you can get the workbook: 

`const workbook = univerInstanceService.getUnit<Workbook>(unitId, UniverInstanceType.UNIVER_SHEET)`

kenny-not-dead:
I understand. I'm not asking you to disclose the paid part of the code, since I've already covered everything I need in terms of business logic... however, as I said, it seemed strange to me that absolutely all events, except this one, do not require intervention on my part, but this still requires it.

Yes, I understand your advice and its implementation, and it will indeed work, but as I said, I think it could work more optimized and not require any additional actions. In any case, it's up to you to decide.

kenny-not-dead:
Can I close this PR or does it take some time?

wpxp123456:
Yes, close it.

PR #6034, opened 2025-10-24
fix: tooltip if content overflow show ellipsis

kenny-not-dead:
close #xxx

When examining link tooltips in documents, if the link text exceeds the tooltip length, the text extends beyond it. As it turns out, this is typical not only for documents but also for tables. I believe the fix in this PR addresses all possible overflow cases.

I'd like to point out that I initially wanted to suggest ellipsis, but I noticed that some links behave differently. There are links that only display on one line if they've been split. Therefore, given that some links can be quite long, it makes sense to impose some kind of line limit.

Before:
image

image

After: 

image

image

## Pull Request Checklist

### DIFF
--- a/packages/design/src/components/tooltip/Tooltip.tsx
+++ b/packages/design/src/components/tooltip/Tooltip.tsx
@@ -254,7 +254,7 @@ export function Tooltip(props: ITooltipProps) {
                 onMouseEnter={() => showTooltip()}
                 onMouseLeave={() => hideTooltip()}
             >
-                <div>{title}</div>
+                <div className="univer-break-words">{title}</div>
                 <div
                     ref={arrowRef}
                     className={`


jikkai:
I think the tooltip should display the full content, but overflow is indeed a problem.

kenny-not-dead:
Please wait a minute, I'll send you an example of a link that's too long, the full length, and it's actually a valid link.

kenny-not-dead:
@jikkai 

Here is an example of a real link, and I think this is not the limit
image

This behavior occurs if you remove the line clamp. If you think it should be removed, I'll fix it. It's up to you to decide.

kenny-not-dead:
But overall, yes, I agree, considering that the tooltip is meant to show content that was already limited, it's strange to cut it off.

jikkai:
Yes, this is exactly the expected behaviour, although it may look a little odd.

---

Actually, Google Sheets doesn’t provide a tooltip for this, so I’m considering whether to remove it. 🤔
For now, let's keep it.

kenny-not-dead:
The tooltip can be a killer feature, but in reality, there are some things that are comparable to Google's, and we find bugs there too.

I removed the line clamp 😊

PR #6074, opened 2025-11-01
fix: thread-comment-ui panel width

kenny-not-dead:
close #xxx
In the comments panel, scrolling appeared along the x-axis. I fixed this by changing the width of this panel.

Additionally, I'd like to discuss another change. I noticed that a significant number of sidebars have a width of 360, while others may have variations of 312, 333, and so on. Do you think it might be worthwhile to create a default value constant for the sidebar, for example, 360, and use it for all sidebars that are smaller than this width, and for those that are larger, set the desired value

The proposal is valid due to the fact that this is essentially a single sidebar, and the user can probably expect a standard size when they appear or change.

If you agree with this statement, I will make a PR about it.

Before:
image
image

After: 
image
image

## Pull Request Checklist

### DIFF
--- a/packages/thread-comment-ui/src/commands/operations/comment.operations.ts
+++ b/packages/thread-comment-ui/src/commands/operations/comment.operations.ts
@@ -34,7 +34,7 @@ export const ToggleSheetCommentPanelOperation: ICommand = {
             sidebarService.open({
                 header: { title: 'threadCommentUI.panel.title' },
                 children: { label: THREAD_COMMENT_PANEL },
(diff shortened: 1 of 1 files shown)


kenny-not-dead:
@jikkai 

Please tell me if this is a problem on my part?
🙂

jikkai:
@weird94 should take a look

kenny-not-dead:
@weird94 
Is this PR still relevant or can it be closed? but you still have scrolling there

PR #6135, opened 2025-11-11
fix: placement at fitst row after entering in the top line

### LINKED ISSUE #6134: [Bug]  unexpected scroll after entering a value

latest

### Reproduction link

I noticed that if I enter values ​​in the very last line, then after pressing enter, I scroll to the very top of the page
If you enter it on the second to last line, this does not happen again.

https://github.com/user-attachments/assets/d1a96785-e46e-4c60-a27f-e4873ed6f9e8

I'm very sorry that lately I've been sending out more bugs than fixes, unfortunately I won't have time(

### Expected behavior

the focus remains where I left off typing

### Actual behavior

 if I enter values ​​in the very last line, then after pressing enter, I scroll to the very top of the page

### System information

_No response_

kenny-not-dead:
close #6134

I still decided to study the problem described in the issues, and I found a solution to this, but perhaps this was the intended solution, so having the same problem with the last column, I would like to offer only line by line for now

I understand that maybe when I'm on the last line and pressing the down arrow on the keyboard, I might expect to go to the first line, but when I'm editing on line 1000, which is quite far away, and pressing enter after entering my value, I don't expect that at all, I'd like to stay at the bottom

I think the user can only expect confirmation of their entered value

By the way, Google Sheets doesn't wrap when saving or when moving with the down arrow. I think it's clear with Excel that there's no limit to the number of rows, so this isn't expected.

from some point of view, one could also consider adding a new line if there is nowhere to go below, this could also be a solution

In any case, it's up to you to decide, but as I noted above, when I first encountered this, I didn't expect it at all

After: 

[Screencast from 2025-11-11 22-20-45.webm](https://github.com/user-attachments/assets/874a91d6-8be6-41c8-8ff8-71e23b403c8e)

## Pull Request Checklist

### DIFF
--- a/packages/sheets-ui/src/commands/commands/utils/selection-utils.ts
+++ b/packages/sheets-ui/src/commands/commands/utils/selection-utils.ts
@@ -84,6 +84,9 @@ export function findNextRange(
             if (next <= boundary.endRow) {
                 destRange.startRow = next;
                 destRange.endRow = next;
+            } else if (next === worksheet.getRowCount()) {
+                destRange.startRow = next - 1;
+                destRange.endRow = next - 1;
             } else if (isGoBack) {
                 destRange.startRow = boundary.startRow;
                 destRange.endRow = boundary.startRow;


wpxp123456:
Thank you very much, but this part shouldn't be modified. We'll only modify the logic for moving the selection after editing the last row/column cell of the worksheet; we'll leave the arrow key selection movement logic unchanged for now.

I submitted this PR: https://github.com/dream-num/univer/pull/6278

PR #6169, opened 2025-11-17
feat: crypto formulas

Dushusir:
## Analysis 

Since similar crypto-related formulas on the market mainly involve pulling encrypted data, they usually require requesting an API. CoinGecko, Binance, Coinbase, and others all have APIs with similar functionality. Users may need to switch between these API addresses and API keys in the Univer to use them. For example, the crypto price query API provides a formula =CRYPTO.PRICE. Internally, it retrieves different data structures based on different API addresses, then performs a transformation layer to convert them into a unified data structure for the formula's result.

## Task

- [ ] The formula plugin supports configuring API addresses and API keys. If not configured, it falls back to https://api.coingecko.com/api/v3/ (Coingecko's public API does not require an API key).
- [ ] Implement the necessary underlying services such as the API request layer and data transformation layer. The actual formula logic should minimize the differences caused by different APIs.
- [ ] First, implement the price query function =CRYPTO.PRICE (referencing Excel notation with the =CRYPTO. prefix), and set several commonly used input parameters (preferably those commonly supported by major exchanges or API providers).

Functions List
- [ ] **Price**
  ```
  =CRYPTO.PRICE("BTC")
  ```
- [ ] **Address Balance**
  ```
  =CRYPTO.BALANCE("0x...", "ethereum")
  ```
- [ ] **Token Balance**
  ```
  =CRYPTO.TOKEN.BALANCE("0x...", "0xTokenAddr", "ethereum")
  ```
- [ ] **Transaction Information**
  ```
  =CRYPTO.TX("ethereum", "0xTxHash")
  ```
- [ ] **NFT Floor Price**
  ```
  =CRYPTO.NFT.FLOOR("azuki")
  ```

## Reference

https://docs.coingecko.com/v3.0.1/reference/simple-price
https://developers.binance.com/docs/binance-spot-api-docs/faqs/market_data_only

## Pull Request Checklist

### DIFF
--- a/packages/engine-formula/src/basics/function.ts
+++ b/packages/engine-formula/src/basics/function.ts
@@ -18,6 +18,7 @@ import type { BooleanNumber } from '@univerjs/core';
 
 import type { FUNCTION_NAMES_ARRAY } from '../functions/array/function-names';
 import type { FUNCTION_NAMES_COMPATIBILITY } from '../functions/compatibility/function-names';
+import type { FUNCTION_NAMES_CRYPTO } from '../functions/crypto/function-names';
 import type { FUNCTION_NAMES_CUBE } from '../functions/cube/function-names';
 import type { FUNCTION_NAMES_DATABASE } from '../functions/database/function-names';
 import type { FUNCTION_NAMES_DATE } from '../functions/date/function-names';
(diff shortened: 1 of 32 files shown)


jikkai (changes requested):
Please create it as a standalone plugin and place it in the packages-experimental directory.

PR #6294, opened 2025-12-12
feat: add page setup precision

### LINKED ISSUE #6212: [Feature] Docs page setup
### Initial checklist

### Problem

Since I have no hope that the next PR can be accepted, I will start a feature

I would like to suggest considering the possibility of removing the long number of characters after the comma (this is a demo), and also considering the issue of limiting the maximum value, since the text is literally lost at high values

Image

for example, when entering 9999 from above. On one of the pages I start to see this (I don't know what it is) just launching a demo

Image

The content itself doesn't even appear on the last page, and as you can see, there isn't even a document border there. This could probably be filed as a separate bug, but I just wanted to point it out.
Image

kenny-not-dead:
close #6212

add precision for indents when setting up a page

After:
image

## Pull Request Checklist

### DIFF
--- a/packages/docs-ui/src/views/page-settings/index.tsx
+++ b/packages/docs-ui/src/views/page-settings/index.tsx
@@ -35,6 +35,10 @@ export interface IPageSettings {
         left: number;
         right: number;
     };
+    pageSise: {
+        width: number;
+        height: number;
+    };
 }
 
 const getPaperSize = (size: ISize) => {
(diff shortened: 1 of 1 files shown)


kenny-not-dead:
I also immediately decided to add the maximum possible value, which I took based on the page size, while considering that you can set the maximum size both at the top and bottom, as well as on the left and right, it seemed to me that you can leave the division by 2 for now, and then think about validation and checking the data when entering.

For example, you could think about calculating the page size from which you need to subtract the mirror indentation, for example, when indenting from the left edge, take the page width and subtract the current right indentation... but these are my thoughts

wpxp123456 (approved):

PR #6306, opened 2025-12-14
fix: defined name input style

### LINKED ISSUE #6305: [Bug]  Long range, overlaps button for the drop-down list

latest

### Reproduction link

The long range overlaps the drop-down list button. This is now noticeable if, for example, you select a range with 2-letter columns and a row greater than 100. but it's also worth noting that this problem will be even more noticeable if there is a specific name there

Image

Image

Image

### Expected behavior

the range does not overlap the drop-down list button

### Actual behavior

Long range, overlaps button for the drop-down list

### System information

_No response_

kenny-not-dead:
close #6305

After:
image

image

it works roughly like in libre
image

but perhaps I would suggest making the input boundaries when focusing, basically like in Google
image

## Pull Request Checklist

### DIFF
--- a/packages/sheets-ui/src/views/defined-name/DefinedName.tsx
+++ b/packages/sheets-ui/src/views/defined-name/DefinedName.tsx
@@ -130,7 +130,7 @@ export function DefinedName({ disable }: { disable: boolean }) {
         >
             <input
                 className={clsx(`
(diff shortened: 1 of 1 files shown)


jikkai:
LGTM

PR #6332, opened 2025-12-17
feat: filter defined name list if  input event

kenny-not-dead:
close #xxx

I will continue to close defined name input resolution issue. 

Considering that there could be quite a few of them and considering the example of Google Sheets, I assumed that it would be a good idea to open a drop-down list when there are matches

Initially, I didn't want to open it if it looked like refString, however, the logic breaks when executing the same test1, so in the end I made this implementation, an example in the video

If you agree with this proposal, perhaps in the next pull request, I would suggest highlighting the matches in the filtered list.

create several defined names and use defined name input

After: 

https://github.com/user-attachments/assets/467cb828-4baf-42c2-a323-fe1647009043

## Pull Request Checklist

### DIFF
--- a/packages/sheets-ui/src/views/defined-name/DefinedName.tsx
+++ b/packages/sheets-ui/src/views/defined-name/DefinedName.tsx
@@ -18,7 +18,7 @@ import type { Workbook, Worksheet } from '@univerjs/core';
 import type { IDefinedNamesServiceParam } from '@univerjs/engine-formula';
 import type { ChangeEvent, KeyboardEvent } from 'react';
 import type { IScrollToCellCommandParams } from '../../commands/commands/set-scroll.command';
-import { AbsoluteRefType, ICommandService, IUniverInstanceService, ThemeService, UniverInstanceType } from '@univerjs/core';
+import { AbsoluteRefType, debounce, ICommandService, IUniverInstanceService, ThemeService, UniverInstanceType } from '@univerjs/core';
 import { borderRightClassName, clsx, Dropdown } from '@univerjs/design';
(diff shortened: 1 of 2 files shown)


wpxp123456:
Good! But it might need some optimization.

1. After  enter or  clicking define name item  and being redirected, you need to close the dropdown menu.
2. The filter should be triggered when the dropdown menu is opened.
3. Is it necessary to use `toLocaleLowerCase` instead of `toLowerCase`?

kenny-not-dead:
@wpxp123456 thanks for watching
1. Yes, I'll watch this episode.
2. I agree, perhaps I was focusing on the case where the range is a valid refString, and in that case, it shouldn't be filtered. I'll look into how to separate these events.
3. my mistake, thank you, I will correct it


I'll fix point 3 today, but points 1 and 2 might only be done by the end of the week, since I plan to be on a business trip tomorrow. I'll report back here once I've fixed them, thanks.

wpxp123456:
Okay, thank you

kenny-not-dead:
@wpxp123456 I've corrected points 1 and 3, but the second point might require a little discussion. I was about to do this, but in the process, I realized the following case: A user, for example, has many defined names, including A1 as DefinedName1 and so when the user is focused on A1, and has DefinedName1 in the top left input, opening the drop-down menu, he will see only it, or the occurrences in the substring, something like DefinedName11 or DefinedName12.
In this particular case, when I simply click on show drop-down list, I as a user would not want filtering, but would like to see everything, for example, to quickly navigate between them, but without having to enter (delete) characters in this input

By the way, this is exactly how Google works. When you simply open it, the filter doesn't work, as you can see from the screenshot.
image

and is triggered when you enter text

image


What do you think about this?

wpxp123456:
Ok

wpxp123456 (approved):

PR #6344, opened 2025-12-18
feat: add remove fonts to facade

kenny-not-dead:
close #xxx

It's clear that your project is popular and used worldwide. Each country has its own linguistic peculiarities, so I would suggest having the most flexible font settings.
For example, out of 15 fonts currently available, only 4 is used most often in my country, and most of the existing ones cannot be used.

Considering the project offers a wide range of customization options for most features, it seems like this wouldn't be a bad idea. I understand that your goal was to prevent the deletion of default fonts, but they could likely be customized by some of your users.

up to you

## Pull Request Checklist

### DIFF
--- a/examples/src/sheets/main.ts
+++ b/examples/src/sheets/main.ts
@@ -200,6 +200,8 @@ function createNewInstance() {
     //     { value: 'Helvetica Neue', label: 'Helvetica Neue', category: 'sans-serif' },
     // ]);
 
+    // window.univerAPI.removeFonts(['SimSun']);
+
     customRegisterEvent(univer, window.univerAPI!);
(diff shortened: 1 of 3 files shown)


jikkai:
It seems a bit strange to dynamically delete fonts through the facade API. I assume what you actually need is to remove those built-in "useless" Chinese fonts :)
If that’s the case, allowing a custom font list to completely override the built-in one would be a more appropriate implementation.

jikkai:
I have opened a new PR that extends the `customFontFamily` configuration, allowing users to completely override the built-in font list. Does this meet your requirements? #6375

kenny-not-dead:
Thanks a lot, it's great)

PR #6346, opened 2025-12-19
Fix Formula, when the cache is setted by STRING, but is get by NUMBER.

wang12124468:
Such as `=SUMIFS(A:A, B:B, "2025")`, when the type of B is STRING (Set cache), the "2025" is transformed by findCompareToken to NUMBER (Get cache)

close #xxx

## Pull Request Checklist


wpxp123456:
Thank you. This problem is known, but it's not possible to simply determine if a number and a string of numbers are equal here, because many formulas use this method.

wpxp123456:
We'll see how to handle this in the future.

PR #6390, opened 2025-12-26
fix: setSelections by top-left input if refString col more 16384

kenny-not-dead:
close #xxx

While thinking about creating defined names through the upper left input, I came across a case where the current method was skipping test1 values ​​and other similar ones. It seems that this method will solve this problem.

Sorry, I didn't immediately think about the need for such a restriction.

## Pull Request Checklist


jikkai (approved):

jikkai:
LGTM 🫡

PR #6468, opened 2026-01-14
feat: use component manager in slider component

kenny-not-dead:
close #xxx

Univer offers great flexibility for customization, including icons. For example, we can easily replace almost all icons in the toolbar. However, in a number of components, such as the slider, this is not implemented, and replacing an icon there in the same way is impossible.

If you support this, I would like to add the same feature to other components where icons cannot be replaced in this way yet.

<!-- BREAKING CHANGE:

Before:
Can't replace the icon in the slider component via the component manager.

After:

Can replace the icon in the slider component via the component manager.
## Pull Request Checklist


kenny-not-dead:
@jikkai  Sorry, were you able to see this? Is a similar implementation possible?

I can also assume that it probably requires a ternary operator? (/ReduceIcon ? /<ReduceIcon />  : null") 
I'll correct it if so

jikkai:
Sorry for the late reply, We have been working on some mobile adaptation recently.

The `ComponentManager` was originally designed to decouple the view layer from the business layer, but it has also brought some extensibility. I agree with your approach of extending icon overrides to more components, and I'm happy to approve your PR.

We will also review the codebase on our side to identify areas where it makes sense to add more UI extension points via the `ComponentManager`, so you won't be doing this alone :)

kenny-not-dead:
@jikkai  Thank you very much for your support and understanding, it's a pleasure working with you

PR #6539, opened 2026-01-30
fix(sheets-ui): recover real value on Microsoft Excel paste with ####…

koushen001:
When copying cells from Microsoft Excel with insufficient column width, Excel's HTML clipboard shows "#######", but the plain text clipboard contains the actual value.

This change uses the plain text data to restore the real number while preserving formats like thousand separators.

- Add plainText parameter to HtmlToUSMService.convert()
- Parse clipboard plain text (TSV format) into 2D grid for cell lookup
- Restore actual value when HTML content displays #######
- Differentiate between genuine "#######" strings and width-truncated display
- Preserve number formatting (thousand separators, decimals, etc.)

close #xxx

## Pull Request Checklist


koushen001:
Test data
After repair
Before repair

PR #6582, opened 2026-02-10
fix(engine-formula): fix incorrect type conversion when comparing str…

klsqmadi:
## Bug Description

When comparing a cell reference containing a date string with a string literal date, the comparison should be lexical (string comparison), not date/numeric comparison.

For example, if cell A1 contains the string "2025-01-01" (as a string literal, not a formatted number), the formula `=A1>="2026-01-01"` should return FALSE (lexical comparison). However, Univer was incorrectly returning TRUE because it was converting the string literal "2026-01-01" to a date number.

close#6392

## Root Cause

The issue was in `ValueObjectFactory.create()` method. When processing string literals like `"2026-01-01"`, it called `stringIsNumberPattern()` which strips the quotes and attempts to parse the content as a date/time/number pattern using `numfmt.parseDate()`. This caused string literals to be incorrectly converted to `NumberValueObject`.

## Fix

Added a check to skip number/date pattern conversion for values wrapped in double quotes (string literals). String literals should always remain as `StringValueObject` to ensure proper lexical comparison.

## Changes

- Modified `ValueObjectFactory.create()` to skip number pattern conversion for quoted strings
- Added test cases for quoted date/time/percentage strings
- Added test cases for string comparison with date literals

## Testing

- All existing tests pass
- New test cases verify the fix works correctly for the reported issue

## Pull Request Checklist


wpxp123456:
image
I'm not sure which version you're using; the latest version has fixed this issue.

It fixed at https://github.com/dream-num/univer/pull/6444/changes

PR #6636, opened 2026-03-03
fix(sidebar): invoke onOpen callback when sidebar is opened

lyr0430:


jikkai:
Thanks for the PR.

It seemed like there were some race conditions, so I have pushed a commit to fix them.

jikkai (approved):

PR #6649, opened 2026-03-05
fix: change font size props value to number; increase font

kenny-not-dead:
close #xxx

I returned to a previously identified issue with increasing the font size after entering it via input. I was even prompted to do this by the fact that the event that generates the font size change was returning a string instead of the expected number. For example:

image

The current solution both fixes the bug reported in #6509 and provides a more expected type for the cell size - number.

image

considering that the IStyleData interface itself defines this as a number

Before:
reported in #6509
After: 

https://github.com/user-attachments/assets/c917a39f-f2f8-4230-bf81-228f0c80f149

## Pull Request Checklist


jikkai (approved):
LGTM

PR #6679, opened 2026-03-13
fix(sheets): icons text and bg not update current color

### LINKED ISSUE #6676: [Bug]  icon's text and background color do not change.

latest

### Reproduction link
Sheets. After we selected a new text or background color, we previously observed that the color of the corresponding icon changed. Afterwards, we could move to any other cell and, for example, apply the last selected color

Now it not only stopped working, but it also doesn't work correctly... the icon's color is always the default, and what's more, it doesn't correspond to what we see after clicking on it again 

https://github.com/user-attachments/assets/a91443aa-2f08-4c6c-8b8c-b28a0f196290

I initially noticed that it might be related to
https://github.com/dream-num/univer/pull/5917

but after researching the issue and going back to the previous implementation I noticed that this does not solve the problem

I was very surprised that not long ago, I made changes to the documents, but it did not affect the icon in the toolbar

After investigating I noticed a difference in useToolbarItemStatus when initializing selectionsValue$, "sheet.command.set-range-text-color" should be an exception

I decided to reopen the issue to offer a possible solution.

### Expected behavior

icon's text and background color do change.

### Actual behavior

icon's text and background color do not change.

### System information

_No response_

kenny-not-dead:
close  #6676

https://github.com/user-attachments/assets/510fe3b5-24f7-4998-9c85-beb91d1d7cd4

After: 

## Pull Request Checklist

### DIFF
--- a/packages/ui/src/views/components/ribbon/hook.ts
+++ b/packages/ui/src/views/components/ribbon/hook.ts
@@ -45,7 +45,7 @@ export function useToolbarItemStatus(menuItem: IDisplayMenuItem<IMenuItem>): ITo
     if (isMenuButtonSelectorItem(menuItem)) {
         const { selections } = menuItem;
 
-        if (Array.isArray(selections)) {
+        if (Array.isArray(selections) && !menuItem.id.startsWith('sheet')) {
             selectionsValue$ = selections?.[0]?.value$;
         }
     }


jikkai:
Thanks for the fix. I opened a [new PR](https://github.com/dream-num/univer/pull/7028) to replace this one.

The issue is not specific to sheet menu ids, so adding a `menuItem.id.startsWith('sheet')` guard in the generic ribbon hook is a bit too narrow. The new PR fixes the shared `BUTTON_SELECTOR` value flow instead, so the toolbar icon uses the latest emitted value while still preserving selection-derived state.

PR #6686, opened 2026-03-14
fix(sheets): update formula refs on clipboard cut/paste

### LINKED ISSUE #6685: [Bug] spreadsheet cut/paste formula reference updates
## Summary
Clipboard cut/paste in sheets does not consistently update formula references the way spreadsheet users expect.

## Current problems
- Cutting a value or range that is referenced by other formulas does not reliably update those dependent references.
- Cutting a formula to a new location can fall through a different path than structural moves, which risks inconsistent reference behavior.
- Cross-sheet cut/paste is especially fragile because move-reference handling can derive sheet identity from the active sheet instead of the actual source/target sheets.
- Mixed and absolute references need Excel-style behavior (`$A$1`, `$A1`, `A$1`).

## Expected behavior
- When a referenced value/range is cut and pasted, formulas that point to it should update to the new address, including range formulas and cross-sheet references.
- Moved formulas should follow Excel move semantics: the moved formula text remains stable, while formulas that reference the moved cell update to the new address.
- Mixed and absolute references should preserve their existing lock behavior.

## Notes
I already have a fix prepared locally that:
- routes clipboard cut/paste through the shared move pipeline
- threads explicit source/target sheet ids through move-effect handling
- prevents double formula rewriting on cut
- adds regression coverage for direct refs, range refs, shared formulas, mixed absolutes, cross-sheet moves, and undo/redo

getmateri:
## Summary
- route spreadsheet clipboard CUT through the shared move pipeline used by structural sheet moves
- thread explicit source and destination sheet ids through move-effect handling so formula, merge, and filter updates work across sheets
- add regression coverage for dependent formulas, mixed absolute refs, shared formulas, cross-sheet moves, and undo/redo

Closes #6685

## Test plan
- [x] `pnpm test -- src/services/clipboard/__tests__/clipboard-service.spec.ts` in `packages/sheets-ui`
- [x] `pnpm test -- src/controllers/__tests__/formula-clipboard.controller.spec.ts` in `packages/sheets-formula-ui`
- [x] `pnpm test -- src/controllers/__tests__/update-formula.controller.spec.ts` in `packages/sheets-formula-ui`
- [x] `pnpm test -- src/controllers/utils/__tests__/ref-range-param.spec.ts` in `packages/sheets-formula`
- [x] `pnpm test -- src/commands/commands/__tests__/move-range-commands.spec.ts` in `packages/sheets`
- [x] `pnpm test -- src/services/__tests__/ref-range.setvice.spec.ts` in `packages/sheets`
- [x] `pnpm test -- src/controllers/__tests__/sheets-filter.controller.spec.ts` in `packages/sheets-filter`


wpxp123456:
@getmateri Could you provide some specific examples of cases that solved which problems?
now, I've only found a problem with the cross-worksheet cut formula.
Furthermore, the movement range should not involve cross-worksheet behavior; I don't understand why you need to know the target worksheet.

getmateri:
@wpxp123456 I added two concrete regression cases here to clarify what this change is covering.

1. Same-sheet CUT/paste of a larger mixed range (values, text, formulas): this is the production issue I hit. Some moved formula refs inside the cut range were being dropped/null after paste, and dependent refs were not updated consistently.
2. Cross-sheet CUT/paste: formulas on the source sheet need to gain the destination sheet qualifier after the move, while formulas on the destination sheet should become local refs once the moved cells now live on that same sheet.

So the target worksheet info is not there to make the low-level move operation broadly cross-worksheet in behavior. It is only threaded into the downstream formula / merge / filter effect handlers so they know where the moved cells land and can serialize references correctly.

I also tightened the formula rewrite so same-workbook moves do not emit workbook-qualified refs, and added explicit coverage for the cross-sheet qualifier rewrite and the larger mixed-range CUT case.

wpxp123456 (approved):

PR #6688, opened 2026-03-14
fix(sheets-ui): prevent arrow key wrap-around at sheet boundaries

### LINKED ISSUE #6687: [Bug] Arrow key navigation wraps around at sheet boundaries instead of stopping
## Description

When pressing arrow keys at the edge of the sheet (e.g., LEFT at column 0, UP at row 0), the cell selector wraps around to the opposite end of the sheet. Both Excel and Google Sheets treat boundaries as hard walls — the selection stays put.

## Steps to Reproduce

1. Open any Univer spreadsheet
2. Click on cell A1 (row 0, column 0)
3. Press the LEFT arrow key

**Expected:** Selection stays at A1
**Actual:** Selection wraps to the last column of the last row

## Root Cause

In `MoveSelectionCommand`, `findNextRange()` is called with its default `isGoBack=true`, which activates wrap-around logic at boundaries.

**File:** `packages/sheets-ui/src/commands/commands/set-selection.command.ts` (line 124)

## Environment

- Univer version: latest dev branch
- Browser: All
- OS: All

getmateri:
## Summary
- Pass `isGoBack=false` to `findNextRange` in `MoveSelectionCommand` so arrow keys stop at sheet edges instead of wrapping to the opposite end, matching Excel and Google Sheets behavior
- Update existing tests to verify boundary clamping instead of wrap-around
- Add dedicated `findNextRange` unit test for `isGoBack=false` at all four edges

Closes #6687

## Test plan
- [x] `selection-utils.spec.ts` — new test verifies `findNextRange` with `isGoBack=false` clamps at all four boundaries
- [x] `set-selections.command.spec.ts` — updated tests verify `MoveSelectionCommand` returns `false` at boundaries and selection stays put
- [x] Tab/Enter wrapping within multi-cell selections is unaffected (separate code path with custom boundary)
- [x] Shift+Arrow expansion already passes `isGoBack=false` — unaffected
- [ ] Manual verification: arrow keys at A1 and at max row/col stay put


wpxp123456:
@getmateri That's how the design works; we only optimized the logic so that pressing the enter key while typing the last row cell doesn't trigger this logic.

getmateri:
@wpxp123456 Interesting -- fair enough. Personally I prefer the hard cell boundaries, but I guess it comes down to preference!

PR #6867, opened 2026-05-06
fix(docs-exchange-ui): dispose previous doc unit on DOCX import

imth:
## Summary

- Repeated DOCX imports kept every prior doc unit alive in the renderer; their headers/footers re-painted at the same page-margin coordinates as the new unit, and the antialiased text edges blended toward white — characters appeared a shade lighter after every import.
- Fix: capture the currently-focused doc unit before creating the new one, focus the new unit, then dispose the previous unit. Creating first avoids a "zero doc unit" frame.

## Test plan

- [x] Import the same DOCX several times in a row in the docs example; header/footer text stays at its original color.
- [x] First-time import (no previous unit) still works — `getCurrentUnitOfType` returns null, dispose is skipped.
- [x] `pnpm test --run` in `packages/docs-exchange` passes (no behavior change in the importer; this PR only touches the UI command).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


imth:
Sorry, opened against the wrong base — meant to PR into my own fork's dev. Closing.

PR #6874, opened 2026-05-07
feat: expose raw editor content in edit lifecycle hooks

### LINKED ISSUE #6863: [Feature] Expose raw editor content in edit lifecycle hooks to support cell custom metadata plugins
### Initial checklist

## Background

I'm building a scientific uncertainty propagation plugin for Univer Sheets. Each cell stores a nominal value in `v` and uncertainty metadata in `custom` (e.g. `{ upperBound: 0.1, upperType: 'abs' }`). The cell displays as `5.00 ± 0.10` via a dynamically generated number format string set by the plugin.

## The Problem

The current edit lifecycle makes it impossible for a plugin to detect when a user has intentionally cleared custom metadata by typing a plain number over an existing uncertainty cell.

Here's what happens when a user double-clicks a cell showing `5 ± 0.1`, clears it, and types `5`:

1. `BEFORE_CELL_EDIT` fires (no raw editor content is available yet)
2. User types `5`, presses Enter
3. Univer merges the `ICellData` derived from the editor document model with the original cell via `Object.assign({}, getCellRaw(), editorCellData)`, so `custom` from the original cell survives into the merged object regardless of what the user typed
4. `diffValue(merged, original)` -> equal -> `SetRangeValuesCommand` is **skipped** (note: the diffValue gate itself is correct behaviour that prevents empty-commit noise; the bug is that the merge runs before the diff, not that the diff exists)
5. `AFTER_CELL_EDIT` fires with the merged data, the plugin cannot distinguish "user typed `5`" from "user typed `5 ± 0.1`"
6. The cell still shows `5 ± 0.1`, the user's intent to clear the uncertainty is silently ignored

The core issue: **`AFTER_CELL_EDIT` receives merged data, not what the user actually typed.** There is no hook between "raw editor output" and "merged result". The merge architecture creates a dead zone where `custom` is effectively immutable through the editor UI: even if the user explicitly overwrites the cell value, the metadata survives.

## What Was Tried

- Nullifying `custom` in `BEFORE_CELL_EDIT` and restoring it on a 500ms timer if Escape was pressed: broke the `±` display inside the editor and is inherently racy
- Listening to DOM input events on the editor element: fragile, not a public API, breaks across refactors
- Monkey-patching the merge function: depends on minified internals, breaks across versions

## Proposed Solution

**Option A (preferred):** Add `rawEditorCellData` to the existing `ISheetLocationForEditor` context passed to `AFTER_CELL_EDIT`. No new interceptor chain, no new API surface, zero breaking changes, just one additional optional field on an existing context object:

```typescript
interface ISheetLocationForEditor {
  // existing fields
  unitId: string;
  subUnitId: string;
  row: number;
  col: number;
  workbook: Workbook;
  worksheet: Worksheet;
  origin: Nullable<ICellData>;
  // proposed addition
  rawEditorCellData?: Nullable<ICellData>; // unmerged, as produced by the document model -> cell conversion
                                           // null when the user clears the editor entirely
}
```

The field is `Nullable` to cover the case where the user clears the editor entirely rather than typing a new value. Plugins should handle both `null` and `undefined` gracefully.

With this, a plugin can check `if (!rawEditorCellData?.custom && origin?.custom)` to detect that the user typed a plain value and dispatch a metadata-clearing mutation.

This is strictly better than the alternatives:

**Option B:** Add an `onBeforeCellCommit` interceptor between editor output and merge. Requires a new interceptor registration point and a new `compose-interceptors` chain, more invasive, more API surface, higher risk of breaking changes.

**Option C:** Expose an `onEditCancel` / `CELL_EDIT_CANCELLED` hook for Escape. Partial solution only, doesn't cover the Enter case where the merged value happens to equal the original.

## Minimal Reproduction

Setting metadata currently requires dispatching a command directly (there is no `setCustom` on the public `FRange` / `FWorksheet` Facade). `unitId` and `subUnitId` can be retrieved via `getSheetCommandTarget()`:

```typescript
import { getSheetCommandTarget } from '@univerjs/sheets';

const target = getSheetCommandTarget(univerInstanceService);
if (!target) return;
const { unitId, subUnitId } = target;

// Set custom metadata on A1
commandService.executeCommand('sheet.command.set-range-custom-metadata', {
  unitId,
  subUnitId,
  range: { startRow: 0, startColumn: 0, endRow: 0, endColumn: 0 },
  custom: { tag: 'important' },
});

// User double-clicks A1, clears content, types "hello", presses Enter
// Expected: custom is cleared, A1 has no metadata
// Actual:   custom still has { tag: 'important' } - merge preserved it, diffValue skipped the command
```

## Why This Matters Beyond This Plugin

Any plugin that stores metadata in `custom` and needs to allow users to clear it through normal editing faces this exact problem. Comments, annotations, data validation metadata, structured data tags: all have the same lifecycle gap. The merge-before-hook architecture makes `custom` effectively append-only through the editor UI. Option A closes that gap with a single optional field and no breaking changes.

https://github.com/user-attachments/assets/1b900924-1575-4671-aba6-f71dd3df113d

CokieMiner:
close #6863

### Description

This PR addresses the "dead zone" in the cell edit lifecycle where custom metadata (stored in the `custom` field of `ICellData`) becomes effectively immutable through the editor UI. 

Currently, `AFTER_CELL_EDIT` receives data that has already been merged with the original cell data. This makes it impossible for plugins to detect when a user intended to clear metadata by overwriting a cell with a plain value, as the original `custom` metadata is preserved during the merge.

**Changes:**
- Added an optional `rawEditorCellData` field to the `ISheetLocationForEditor` interface.
- Updated `SheetInterceptorService.onWriteCell` to accept and propagate this raw data.
- Modified `EditingRenderController` to calculate the "unmerged" cell data (produced directly from the document model conversion) and pass it to the interceptor service.

This implementation allows plugins to distinguish user intent by checking the presence of fields in `rawEditorCellData` vs the merged `cellData`.

### How to test them.

1. Create a plugin that intercepts `AFTER_CELL_EDIT`.
2. Set custom metadata on a cell (e.g., `{ custom: { myMetadata: true } }`).
3. Double-click the cell, clear it, or type a new plain value and press Enter.
4. Verify that `context.rawEditorCellData` correctly reflects the "unmerged" state:
   - If the cell was cleared, `rawEditorCellData` should be `null`.
   - If a plain value was typed, `rawEditorCellData.custom` should be `undefined`, even though the merged `cellData.custom` still contains the original metadata.
5. `pnpm typecheck` has been run and passes across the workspace.

I have manually verified with a scientific uncertainty propagation plugin; the `rawEditorCellData` field correctly enables metadata-clearing logic. [Example](https://github.com/CokieMiner/Anafis-Tauri/commit/ed4ece7ccf44a2cd5841b49e46f097e6269a1be5)

## Pull Request Checklist

- [x] Related tickets or issues have been linked in the PR description (close #6863).


wpxp123456:
This change is unnecessary. If you just want to know if the value has changed after editing, you can simply compare the `cellData` with the corresponding cell data in the `worksheet`, and then customize your return value.

wpxp123456:
We'll turn this off for now.

CokieMiner:
You didn't read the original problem, but it's ok I found a way of having this without having to have the whole repo inside my own project, by using a automatic patcher, thanks for your time.

PR #6890, opened 2026-05-12
fix: remove scroll bar on sheets context menu

kenny-not-dead:
close #xxx

We noticed an unnecessary scroll when trying to change the color of the sheet, but I couldn't find any places where this part is also called, other than this feature.

At the same time, it seems we have not observed such a problem before, but I have not found a reason that could lead to this

Before:
image

After: 
image

## Pull Request Checklist


jikkai:
Thanks for the PR! I've identified this as a CSS box model issue and have submitted a fix in #6892

PR #6910, opened 2026-05-16
fix(sheets): paste images as cell images while editing

IMSupperkaka:
## What

Fixes image paste behavior in sheet cell editing mode. When the user pastes an image while a cell editor is active, the image now follows the same path as inserting a cell image instead of being embedded into the editor content as base64.

## Why

Previously, clipboard image files pasted during cell editing were handled by the docs editor paste flow. This could embed the image as base64 inside the cell content, increasing workbook size and differing from the normal "insert cell image" behavior. In some cases the same paste could also continue through the sheet file paste hook and insert a floating image, causing duplicate insertion.

## How

- Route image files pasted in sheet editor units through `InsertCellImageCommand` with the clipboard files.
- Let `InsertCellImageCommand` accept file parameters and reuse the existing `insertCellImageByFile` upload path.
- Allow docs image paste hooks to consume an image by returning `null`, preventing base64 inline insertion.
- Suppress the floating image paste hook for the same paste event when the edit-mode cell image path has already consumed the image.
- Preserve the existing non-editing sheet paste behavior, which continues to insert floating images.

## Validation

- Ran targeted eslint on the modified files. Existing warnings remain in `packages/docs-ui/src/services/clipboard/clipboard.service.ts`; no new errors were introduced.

### DIFF
--- a/packages/docs-ui/src/services/clipboard/clipboard.service.ts
+++ b/packages/docs-ui/src/services/clipboard/clipboard.service.ts
@@ -35,6 +35,7 @@ import {
     ObjectRelativeFromH,
     ObjectRelativeFromV,
     PositionedObjectLayoutType,
+    SHEET_EDITOR_UNITS,
     SliceBodyType,
     toDisposable,
     Tools,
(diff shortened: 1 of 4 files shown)


wpxp123456:
Thanks, the implementation in this PR should also solve your issue, so I’ll close your PR for now.

https://github.com/dream-num/univer/pull/6985

IMSupperkaka:
> Thanks, the implementation in this PR should also solve your issue, so I’ll close your PR for now.
> 
> #6985

It seems that they are not solving the same problem. #6985 This PR solves the problem of parsing the base64 format pictures in the pasteboard. My PR is mainly to solve the problem of pasting pictures in the active cell without using IImageIoService. This will cause the pictures to be stored in the table data in base64 format.

PR #6933, opened 2026-05-21
feat(sheets): add R1C1 formula facade API

hexf00:
## Summary

- Add `normalizeFormulaR1C1ToA1()` in `@univerjs/engine-formula` to normalize R1C1 references in a formula string into A1 references for a target cell.
- Improve `deserializeRangeForR1C1()` so common R1C1 forms such as `RC`, `R1C`, `RC1`, and relative references like `R[1]C[-1]` parse correctly without treating `[1]` as a workbook id.
- Add `FRange.setFormulaR1C1(formula)` and `FRange.setFormulasR1C1(formulas)`, resolving relative references against each target cell.

## Design Notes

This PR intentionally implements R1C1 at the `FRange` formula layer first. Once the RangeList API PR lands, `FRangeList.setFormulaR1C1()` and `FRangeList.setFormulasR1C1()` can safely delegate to these `FRange` APIs without inventing a separate execution path.

The normalizer skips double-quoted string literals, skips R1C1-looking text inside structured-reference brackets, and supports sheet/workbook-qualified R1C1 references. It normalizes to A1 because the current Univer facade formula write path already accepts A1 formulas.

This is not a full parser-level R1C1 reference-style mode. A future formula-engine PR can move this lower into lexer/reference-token handling if we decide to support R1C1 natively in the AST pipeline.

## Tests

- `pnpm --filter @univerjs/engine-formula test src/engine/utils/__tests__/r1c1-reference.spec.ts`
- `pnpm --filter @univerjs/sheets test src/facade/__tests__/f-range.spec.ts`
- `pnpm --filter @univerjs/sheets typecheck`
- `pnpm exec eslint packages/engine-formula/src/engine/utils/r1c1-reference.ts packages/engine-formula/src/engine/utils/__tests__/r1c1-reference.spec.ts packages/engine-formula/src/index.ts packages/sheets/src/facade/f-range.ts packages/sheets/src/facade/__tests__/f-range.spec.ts`
- `git diff --check`

Note: targeted eslint exits with 0 and reports only existing warnings in `packages/sheets/src/facade/f-range.ts` / `f-range.spec.ts`.

### DIFF
--- a/packages/engine-formula/src/engine/utils/r1c1-reference.ts
+++ b/packages/engine-formula/src/engine/utils/r1c1-reference.ts
@@ -16,23 +16,103 @@
 
 import type { IRange, IUnitRangeName } from '@univerjs/core';
 import { AbsoluteRefType } from '@univerjs/core';
-import { handleRefStringInfo } from './reference';
+import { serializeRangeToRefString, unquoteSheetName } from './reference';
 
-const $relativeRegex = /[\[\]]/g;
+const $relativeRegex = /[\[\]]/;
+const $relativeReplaceRegex = /[\[\]]/g;
+const $formulaR1C1ReferenceRegex = /(?:(?:'(?:(?:'')|[^'])+'|(?:\[[^\]\r\n]+\])?[A-Za-z_][A-Za-z0-9_.]*)!)?R(?:\[-?\d+\]|\d*)C(?:\[-?\d+\]|\d*)(?::R(?:\[-?\d+\]|\d*)C(?:\[-?\d+\]|\d*))?/iy;
+const $workbookSheetPrefixRegex = /^\[([^\]\r\n]+)\](.*)$/;
 
 function handleR1C1(rowOrColumnString: string, current: number) {
+    if (rowOrColumnString === '') {
+        return current;
+    }
+
     if ($relativeRegex.test(rowOrColumnString)) {
-        const index = Number(rowOrColumnString.replace($relativeRegex, ''));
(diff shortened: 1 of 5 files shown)


packages/engine-formula/src/engine/utils/r1c1-reference.ts
```diff
@@ -123,6 +203,103 @@ export function deserializeRangeForR1C1(refString: string, currentRow = 0, curre
...
+        const char = formulaString[index];
+
+        if (char === '"') {
+            const endIndex = readDoubleQuotedString(formulaString, index);
+            result += formulaString.slice(index, endIndex);
+            index = endIndex;
+            continue;
+        }
+
+        const matched = formulaString.slice(index).match($formulaR1C1ReferenceRegex);
```
jikkai: slice + match in a loop is O(n²) for long formulas

jikkai (changes requested):

packages/engine-formula/src/engine/utils/r1c1-reference.ts
hexf00: Updated. The normalizer now uses a sticky regex with `lastIndex` and `exec(formulaString)` instead of `formulaString.slice(index).match(...)`, and accumulates output in chunks before `join()`, so the scan avoids repeated substring allocation for long formulas.

jikkai (approved):
LGTM

hexf00:
Not ready to merge yet

PR #6943, opened 2026-05-23
fix(engine-formula): resolve Excel compatibility function prefixes

### LINKED ISSUE #4815: [Feature] Support future function call with "_xlfn." prefix in formula for compatibility
### Initial checklist

### Problem

Univer does not correctly recognize the "_xlfn." prefix in formulas, even when the corresponding function exists in Univer. For example, in the formula `"_xlfn.STDEV.S(Returns!C3:C253)"`, the function `"STDEV.S"` is already available in Univer. The "_xlfn." prefix indicates that the formula originates from a newer version of Excel and that the function may not exist in the current version. However, since the function is present in Univer, the prefix should be handled appropriately.

For more info about "_xlfn.":
https://libxlsxwriter.github.io/working_with_formulas.html
https://learn.microsoft.com/en-us/openspecs/office_standards/ms-xlsx/5d1b6d44-6fc1-4ecd-8fef-0b27406cc2bf?redirectedfrom=MSDN

IMSupperkaka:
## What
- Resolve registered functions even when imported formulas include Excel compatibility prefixes such as _xlfn.
- Apply the same lookup behavior to function descriptions.
- Add service tests for prefixed executor and description lookups.

- Closes #4815

- ## Test
- pnpm --filter @univerjs/engine-formula test -- src/services/__tests__/function.service.spec.ts

### DIFF
--- a/packages/engine-formula/src/services/__tests__/function.service.spec.ts
+++ b/packages/engine-formula/src/services/__tests__/function.service.spec.ts
@@ -35,6 +35,16 @@ describe('FunctionService', () => {
         expect(service.hasExecutor('AVERAGE')).toBe(false);
     });
 
+    it('should resolve Excel compatibility prefixed executors', () => {
+        const service = new FunctionService();
+        const stdevExecutor = { name: 'STDEV.S' };
+
+        service.registerExecutors(stdevExecutor as never);
+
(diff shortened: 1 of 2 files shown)


wpxp123456:
This needs to be removed at the data layer, not handled inside Univer. If you encountered this issue while using our import/export feature, please provide the Excel file.

PR #6945, opened 2026-05-23
fix(engine-formula): keep quoted dates as strings

### LINKED ISSUE #6392: [Bug] Incorrect type conversion while comparing a cell reference containing date as string literal and another string literal

0.10.14, 0.12.4, 0.14.0

### Reproduction link

https://stackblitz.com/edit/univer-sheets-vite-cfpfg2nd?file=src%2Fmain.ts

### Expected behavior

When comparing a cell reference containing "2025-01-01" as string literal(not a formatted number), and another string literal "2026-01-01"

=A1>="2026-01-01"

The expected output is FALSE,

### Actual behavior

Univer always returns TRUE in such cases.

Diving deeper into the code it looks like Univer tries to interpret the RHS i.e "2026-01-01" as date, and hence while doing comparison does a comparison like

"2025-01-01" >= 4555

Whenever a string is compared against a number with >=, univer formula engine always returns TRUE.

This behavior is completely different from MS Excel, and Google Sheet, where they seem to do lexical comparison of the LHS and RHS.

Screenshot 1 -> Incorrect formula output
Image

Screenshot 2 -> Correct formula output when TEXT() function is used to skip automatic type conversion
Image

Screenshot 3 -> Correct formula output when TEXT() function is used to skip automatic type conversion
Image

### System information

_No response_

IMSupperkaka:
## What
- Do not parse explicitly quoted formula literals as date/number formats.
- Preserve lexical comparison for cell strings compared with quoted date strings.
- Add parser and value-object regression tests for quoted date strings.

- Closes #6392

- ## Test
- pnpm --filter @univerjs/engine-formula test -- src/engine/analysis/__tests__/parser.spec.ts src/engine/value-object/__tests__/array-value-object.spec.ts

### DIFF
--- a/packages/engine-formula/src/engine/analysis/__tests__/create-command-test-bed.ts
+++ b/packages/engine-formula/src/engine/analysis/__tests__/create-command-test-bed.ts
@@ -68,6 +68,8 @@ const TEST_WORKBOOK_DATA: IWorkbookData = {
         sheet1: {
             id: 'sheet1',
             name: 'Main',
+            rowCount: 5,
+            columnCount: 5,
             cellData: {
                 0: {
                     0: {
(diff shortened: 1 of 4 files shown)


wpxp123456:
This issue has been fixed in the latest code.

PR #6983, opened 2026-05-29
fix(engine-formula): skip embedded refs for other formulas

### LINKED ISSUE #6984: bug(engine-formula): other formula embedded array refs leak into worksheet snapshot
## Problem

When array formula snapshot writing is enabled, formulas registered by non-worksheet features such as conditional formatting can record embedded array refs as worksheet cell metadata.

This can pollute workbook snapshot cellData with refs that do not belong to normal worksheet cell formulas. In downstream xlsx export paths, this was observed as worksheet sheetData being exported without cell nodes after import/export roundtrip.

Related discovery issue: dream-num/univer-cli#320.

## Concrete Example

A workbook has a conditional formatting rule applied to `B1:B5` with a custom formula like:

```text
=IF(A1:A5, "open")
```

That formula can return an array result while the condition is evaluated for the formatting rule. It is not a worksheet cell formula; it is registered by conditional formatting as an other formula and its value should be consumed by conditional formatting through `unitOtherData`.

With `writeArrayFormulaToSnapshot: true`, the array result can currently call `setUnitArrayFormulaEmbeddedMap()`. Because the runtime does not know the current formula source, the conditional-formatting formula can be recorded as worksheet embedded array-ref metadata, even though no real worksheet cell formula owns that ref.

## Root Cause

Formula values are already separated by source:

- normal worksheet formulas write worksheet runtime data
- other formulas write unitOtherData
- feature formulas write feature runtime data

However, embedded array refs were collected from AST execution through FormulaRuntimeService without checking the current formula source. Conditional formatting formulas are registered as other formulas, but their array results could still call setUnitArrayFormulaEmbeddedMap and enter worksheet-only array formula metadata.

## Expected Behavior

Embedded array refs should only be collected for normal worksheet cell formulas. Other formula and feature formula results should remain in their corresponding runtime data paths and must not write worksheet array formula metadata.

## Fix Direction

Track the current FormulaDependencyTreeType in FormulaRuntimeService and only collect embedded array refs when the current formula type is NORMAL_FORMULA.

OTime-Coder:
## Summary

- Track the current dependency tree formula type in formula runtime context.
- Only collect embedded array formula refs for normal worksheet formulas.
- Pass formula dependency type through calculation and dependency analysis paths.
- Add a runtime regression test for OTHER_FORMULA vs NORMAL_FORMULA embedded ref collection.

Closes #6984

## Validation

- pnpm exec vitest run packages/engine-formula/src/services/__tests__/runtime.service.spec.ts
- pnpm --filter @univerjs/engine-formula typecheck

### DIFF
--- a/packages/engine-formula/src/engine/dependency/formula-dependency.ts
+++ b/packages/engine-formula/src/engine/dependency/formula-dependency.ts
@@ -720,7 +720,8 @@ export class FormulaDependencyGenerator extends Disposable implements IFormulaDe
                 tree.rowCount,
                 tree.columnCount,
                 tree.subUnitId,
-                tree.unitId
+                tree.unitId,
+                tree.type
             );
 
             const rangeList = await this._getRangeListByNode({
@@ -1039,7 +1040,8 @@ export class FormulaDependencyGenerator extends Disposable implements IFormulaDe
             tree.rowCount,
             tree.columnCount,
             tree.subUnitId,
-            tree.unitId
+            tree.unitId,
+            tree.type
         );
 
         const dirtyRanges: IUnitRange[] = [];
@@ -1092,7 +1094,8 @@ export class FormulaDependencyGenerator extends Disposable implements IFormulaDe
                 tree.rowCount,
                 tree.columnCount,
                 tree.subUnitId,
-                tree.unitId
+                tree.unitId,
+                tree.type
             );
 
             let value: FunctionVariantType;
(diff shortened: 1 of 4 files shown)


wpxp123456:
Thanks for reporting this. This is no longer reproducible on the current `dev` branch.

The `writeArrayFormulaToSnapshot` option and the snapshot-writing path were removed in #7267. Other formula results are now kept in `unitOtherData` and are not written to worksheet snapshots.

PR #7074, opened 2026-06-12
fix(docs): initialize body for partial document data

hexf00:
Related downstream report: Univer CLI document creation followed by Facade text insertion failed because the created document had no body.

This PR makes partial document data without `body` initialize as a standard empty document in `DocumentDataModel`. This is necessary because Facade callers can create documents with only `id` and `title`, then naturally call `appendText` or `insertText`; before this change the document had no body and those APIs failed with `The document body is empty`.

The fix keeps a single document creation path by normalizing missing-body document snapshots at the core data-model boundary. It does not add CLI string matching or Facade fallback behavior. Documents that already provide a body keep their supplied body.

How to test them:

- `pnpm --filter @univerjs/core exec vitest run src/docs/data-model/__tests__/document-data-model.spec.ts src/docs/data-model/__tests__/empty-snapshot.spec.ts`
- `pnpm --filter @univerjs/docs exec vitest run src/facade/__tests__/f-document-node.spec.ts`
- `pnpm --filter @univerjs/core typecheck`
- `pnpm --filter @univerjs/docs typecheck`
- `git diff --check`

## Pull Request Checklist

- [x] Related downstream report is described in the PR description.
- [x] Naming convention is followed.
- [x] Unit tests have been added for the changes.
- [x] No breaking changes introduced in this PR.

### DIFF
--- a/packages/core/src/docs/data-model/__tests__/document-data-model.spec.ts
+++ b/packages/core/src/docs/data-model/__tests__/document-data-model.spec.ts
@@ -1,0 +1,36 @@
+/**
+ * Copyright 2023-present DreamNum Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { describe, expect, it } from 'vitest';
+import { DocumentDataModel } from '../document-data-model';
+
+describe('DocumentDataModel', () => {
+    it('normalizes document data without body to an editable empty document', () => {
+        const document = new DocumentDataModel({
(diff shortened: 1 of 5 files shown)


jikkai:
What is the original problem, and is there a way to reproduce it?

hexf00:
Original downstream issue: https://github.com/dream-num/univer-cli/issues/461

The problem was observed through Univer CLI SaC document authoring. Users could discover `univerAPI.createDocument`, but there was no safe path for constructing a legal document body. The natural Facade usage is:

```ts
const fDocument = univerAPI.createDocument({
  id: 'doc-demo',
  title: 'Demo Doc',
  documentStyle: {},
});

fDocument.appendText('Univer');
// The downstream report used insertParagraph after createDocument in the same missing-body path.
```

Before this PR, that partial document snapshot kept `body` missing, so the first text insertion failed with `The document body is empty`.

The downstream report also contains a second symptom: after users tried to work around the error by hand-writing a minimal body, the handcrafted body had invalid/incomplete document structure and two paragraph insertions collapsed into one paragraph. That is why the downstream issue looks like several separate gaps. The root blocker for this PR is the missing-body partial snapshot: callers should not need to handcraft `dataStream` / `paragraphs` / `sectionBreaks` just to create an editable empty document.

Expected behavior after this PR: `DocumentDataModel` normalizes document data without `body` to the standard empty document body (`dataStream: '\r\n'`, one paragraph at `0`, one section break at `1`), and Facade text insertion works from there. Caller-supplied malformed bodies are not the target of this PR.

PR #7150, opened 2026-06-24
feat(board): support explicit locale exports

DR-Univer:
## Summary

- add explicit locale publish exports for packages that opt out of wildcard exports
- keep wildcard publish behavior unchanged by default
- improve design Button `asChild` typing and Switch accessibility semantics for board UI controls

## Why

The board packages in pro need stable explicit locale entrypoints during bundling and publishing. Board settings also rely on Univer design controls behaving like accessible switches while preserving existing visual behavior.

## Validation

- `pnpm --filter @univerjs/design exec vitest run src/components/switch/__tests__/index.spec.tsx --reporter=verbose`
- `pnpm exec vitest run common/shared/tsdown/utils/cleanup-pkg.spec.ts --reporter=verbose`

### DIFF
--- a/packages/design/src/components/button/Button.tsx
+++ b/packages/design/src/components/button/Button.tsx
@@ -15,7 +15,7 @@
  */
 
 import type { VariantProps } from 'class-variance-authority';
-import type { ButtonHTMLAttributes } from 'react';
+import type { ButtonHTMLAttributes, ComponentPropsWithoutRef } from 'react';
 import { Slot } from '@radix-ui/react-slot';
 import { cva } from 'class-variance-authority';
 import { forwardRef } from 'react';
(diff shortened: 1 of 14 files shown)


packages/design/src/components/button/Button.tsx
```diff
@@ -94,10 +94,24 @@ export interface IButtonProps
 
 export const Button = forwardRef<HTMLButtonElement, IButtonProps>(
     ({ className, variant, size, asChild = false, ...props }, ref) => {
-        const Comp = asChild ? Slot : 'button';
+        const buttonClassName = clsx(buttonVariants({ variant, size, className }));
```
jikkai: Do not modify the base components.

jikkai (changes requested):

packages/design/src/components/switch/Switch.tsx
```diff
@@ -14,16 +14,18 @@
  * limitations under the License.
  */
 
+import type { InputHTMLAttributes } from 'react';
```
jikkai: Do not modify the base components.

jikkai (changes requested):

PR #7182, opened 2026-07-03
feat(board): support explicit locale exports

DR-Univer:
## Summary

- add explicit locale publish exports for packages that opt out of wildcard exports
- keep wildcard publish behavior unchanged by default
- expose no-wrap text measurement support needed by board text sizing

## Why

The board packages in pro need stable explicit locale entrypoints during bundling and publishing, and board shape text sizing needs a reusable no-wrap measurement path from engine-render.

## Validation

- `./node_modules/.bin/vitest run src/components/docs/layout/__tests__/doc-no-wrap-measure.spec.ts`


packages/design/src/components/switch/Switch.tsx
```diff
@@ -14,16 +14,18 @@
  * limitations under the License.
  */
 
+import type { InputHTMLAttributes } from 'react';
```
jikkai: Invalid changes

common/shared/tsdown/utils/cleanup-pkg.ts
```diff
@@ -20,6 +20,9 @@ type CleanupPackageJson = IPackageJson & {
         main: string;
         module: string;
     };
+    univerCli?: {
```
jikkai: ?

common/shared/tsdown/utils/cleanup-pkg.ts
```diff
@@ -357,8 +360,16 @@ function deriveDependencyGroups(packageDir: string, packageJson: IPackageJson):
...
+    const localeDir = path.resolve(packageDir, 'src/locale');
+    const localeNames = fs.existsSync(localeDir)
+        ? fs.readdirSync(localeDir)
+            .filter((fileName) => path.extname(fileName) === '.ts')
+            .map((fileName) => path.basename(fileName, '.ts'))
+            .sort((left, right) => left.localeCompare(right))
+        : [];
+    const hasLocales = localeNames.length > 0;
     const hasFacade = fs.existsSync(path.resolve(packageDir, 'src/facade/index.ts'));
+    const publishWildcardExports = pkg.univerCli?.publishWildcardExports !== false;
```
jikkai: FOR WHAT?????

jikkai (changes requested):

PR #7212, opened 2026-07-07
fix(engine-formula): improve formula compatibility stability

DR-Univer:
## Summary
- Improve engine-formula compatibility and formula evaluation stability across parser/interpreter/value/reference handling.
- Add and update regression coverage for formula functions, shared formulas, dynamic references, formula data, runtime behavior, and new Excel-compatible functions.
- Keep the branch rebased on latest origin/dev.

## Validation
- node_modules/.bin/tsc.cmd --noEmit -p packages/engine-formula/tsconfig.json
- node vitest.mjs run targeted engine-formula specs: cell-reference-object, indirect, sumproduct, formula-data.model, runtime.service, calculate-formula.service

Note: pnpm test could not be used directly because current lockfile policy rejects recent packages under minimumReleaseAge; tests were run via the existing vitest binary to avoid invoking pnpm install.


packages/engine-formula/src/services/defined-names.service.ts
```diff
@@ -224,27 +224,60 @@ export class DefinedNamesService extends Disposable implements IDefinedNamesServ
...
+
+        if (sheetId) {
+            const normalizedName = name.toLowerCase();
+            let workbookScope: Nullable<IDefinedNamesServiceParam> = null;
+            let firstMatch: Nullable<IDefinedNamesServiceParam> = null;
+            for (const item of Object.values(nameMap)) {
+                if (item.name.toLowerCase() !== normalizedName) {
+                    continue;
+                }
+                firstMatch ??= item;
```
wpxp123456: ??= 有兼容性问题

pnpm-workspace.yaml
```diff
@@ -14,3 +14,8 @@ overrides:
     basic-ftp: 5.2.0
     react: 19.2.7
     react-dom: 19.2.7
+allowBuilds:
```
wpxp123456: pnpm-workspace 的改动删掉

PR #7240, opened 2026-07-11
fix(engine-render): preserve fractional document font sizes

realmorrisliu:
## Summary
- preserve positive finite fractional font sizes instead of rounding them up
- reject zero, negative, NaN, and infinite font sizes and fall back to the default
- add regression coverage for 8pt, 9.2pt, and 10.4pt document typography

## Validation
- pnpm --filter @univerjs/engine-render exec vitest run src/basics/__tests__/tools.spec.ts
- pnpm --filter @univerjs/engine-render typecheck
- pnpm exec eslint packages/engine-render/src/basics/tools.ts packages/engine-render/src/basics/__tests__/tools.spec.ts (0 errors; existing warnings only)

### DIFF
--- a/packages/engine-render/src/basics/__tests__/tools.spec.ts
+++ b/packages/engine-render/src/basics/__tests__/tools.spec.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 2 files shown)


realmorrisliu:
Closing at the author's request. The renderer investigation will remain local and will not be proposed to the Univer repository through this PR.

PR #7241, opened 2026-07-11
fix(sheets-drawing-ui): guard getAllFloatDoms against float doms without a backing drawing param

### LINKED ISSUE #7132: [Bug] getAllFloatDoms api call error

0.25.0

### Reproduction link

https://github.com/QJerryxjh/test_univer/tree/getAllFloatDom

### Expected behavior

call getAllFloatDom can get all float dom, work success

### Actual behavior

Image

### System information

 System:
    OS: macOS 26.5.1
    CPU: (8) arm64 Apple M1
    Memory: 265.42 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.10.0 - /Users/qjerry/.nvm/versions/node/v20.10.0/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 10.2.3 - /Users/qjerry/.nvm/versions/node/v20.10.0/bin/npm
    pnpm: 10.5.2 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 149.0.7827.116
    Firefox: 151.0.4
    Safari: 26.5

mvanhorn:
Closes #7132

## Summary

`FWorksheet.getAllFloatDoms()` crashed with a TypeError when a worksheet contained a float dom that the drawing manager tracked but for which `getDrawingByParam(...)` could not resolve a backing drawing param. The method non-null-asserted that lookup and then read `.componentKey`/`.data` off the result inside `.map()`, so a single unresolvable entry threw and took down the entire call, leaving `getAllFloatDoms` unusable for the whole sheet.

## Why

The sibling accessor `getFloatDomById()` already guards the exact same `getDrawingByParam(...)` lookup with an early `if (!drawingParm) return null;`, so the two paths disagreed: fetching one float dom degraded gracefully while fetching all of them threw. That inconsistency is the reported crash. The fix brings `getAllFloatDoms()` in line with its sibling: skip entries whose drawing param cannot be resolved, then filter the resulting nulls out with a type guard so the method's return type is preserved and callers still get a clean array of resolvable float doms.

## How to test

Added a regression test in `packages/sheets-drawing-ui/src/facade/__tests__/f-worksheet.spec.ts` ("omits float doms without backing drawing params") that registers a float dom whose backing drawing param is absent and asserts `getAllFloatDoms()` returns the resolvable entries without throwing. Run it with:

```
npx vitest run packages/sheets-drawing-ui/src/facade/__tests__/f-worksheet.spec.ts -t "omits float doms without backing drawing params"
```

## Pull Request Checklist

### DIFF
--- a/packages/sheets-drawing-ui/src/facade/__tests__/f-worksheet.spec.ts
+++ b/packages/sheets-drawing-ui/src/facade/__tests__/f-worksheet.spec.ts
@@ -547,8 +547,14 @@ describe('sheets-drawing-ui facade', () => {
 
         const worksheet = univerAPI.getActiveWorkbook()!.getActiveSheet();
         const sheetDrawingService = injector.get(ISheetDrawingService);
-        worksheet.addFloatDomToPosition({ componentKey: 'OrderCard' }, 'order-card');
-        worksheet.addFloatDomToPosition({ componentKey: 'MissingCard' }, 'missing-card');
+        worksheet.addFloatDomToPosition({
+            componentKey: 'OrderCard',
+            initPosition: { startX: 20, startY: 30, endX: 140, endY: 90 },
+        }, 'order-card');
+        worksheet.addFloatDomToPosition({
+            componentKey: 'MissingCard',
+            initPosition: { startX: 40, startY: 50, endX: 160, endY: 110 },
+        }, 'missing-card');
         await Promise.resolve();
 
         const drawingData = sheetDrawingService.getDrawingData('test', 'sheet1');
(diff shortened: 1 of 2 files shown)


mvanhorn:
Fixed the typecheck failure in d3de569: the map callback needed an explicit `IFCanvasFloatDomResult | null` return type for the null guard. Package build, typecheck, lint, and the 129 package tests pass.

wpxp123456 (approved):

PR #7249, opened 2026-07-13
feat(ui): make FloatDom content insets configurable

siam-ese:
## What changed

- Add a flexible `contentBox` configuration to FloatDom with independent `wrapperInset` and `contentInset` values.
- Preserve the historical wrapper/content defaults of `2 / 4` when the config or an individual field is omitted.
- Support exact placement and content bounds through `{ wrapperInset: 0, contentInset: 0 }`.
- Keep configured values as caller-owned external behavior without runtime validation or correction.
- Add unit and browser coverage for defaults, partial/custom values, clipping, zoom, rotation, borders, resize, and rotate interactions.

## Why

The previous closed mode API only represented two layouts. Numeric insets expose the actual layout behavior and let consumers choose intermediate or custom values without adding more modes.

## Compatibility and impact

Existing FloatDom consumers are unchanged because omitted configuration retains the historical `2 / 4` behavior.

Persisted command or mutation schemas are unchanged. No historical payload decoder or migration is required because this is a runtime FloatDom layer configuration, not a persisted command/mutation contract.

## Validation

- UI Vitest: 63 files, 279 tests passed
- UI typecheck passed
- Examples typecheck and e2e build passed
- Focused ESLint: 0 errors
- Playwright: 2 tests passed, including the 160-case browser matrix and resize/rotate interaction coverage
- `git diff --check` passed

## Pull Request Checklist

- [x] [Naming convention](https://github.com/dream-num/univer/blob/dev/docs/NAMING_CONVENTION.md) is followed.
- [x] Unit tests have been added for the changes.


packages/ui/src/views/components/dom/FloatDom.tsx
```diff
@@ -18,84 +18,66 @@ import type { CSSProperties } from 'react';
...
+    wrapper: HTMLDivElement,
+    inner: HTMLDivElement,
+    layout: ReturnType<typeof resolveFloatDomLayout>
+): void {
+    const { wrapper: wrapperStyle, inner: innerStyle } = layout;
+    setStyleProperty(wrapper.style, 'top', `${wrapperStyle.top}px`);
+    setStyleProperty(wrapper.style, 'left', `${wrapperStyle.left}px`);
+    setStyleProperty(wrapper.style, 'width', `${wrapperStyle.width}px`);
+    setStyleProperty(wrapper.style, 'height', `${wrapperStyle.height}px`);
+    setStyleProperty(wrapper.style, 'transform', wrapperStyle.transform);
```
jikkai: Why not use the clsx + Tailwind CSS system to modify static styles?

packages/ui/src/views/components/dom/FloatDom.tsx
```diff
@@ -18,84 +18,66 @@ import type { CSSProperties } from 'react';
...
 import { DocumentDataModel, IUniverInstanceService } from '@univerjs/core';
 import { memo, useEffect, useMemo, useRef } from 'react';
-import { distinctUntilChanged, first } from 'rxjs';
+import { first } from 'rxjs';
 import { ComponentManager } from '../../../common';
 import { CanvasFloatDomService, shouldForwardFloatDomEvents, shouldRenderFloatDomLayer } from '../../../services/dom/canvas-dom-layer.service';
 import { useDependency, useObservable } from '../../../utils/di';
+import { resolveFloatDomLayout } from './float-dom-layout';
+
+function setStyleProperty(style: CSSStyleDeclaration, property: string, value: string): void {
```
jikkai: Does this encapsulation make any sense? There would be no need for attribute defense here at all

packages/ui/src/views/components/dom/FloatDom.tsx
```diff
@@ -18,84 +18,66 @@ import type { CSSProperties } from 'react';
...
+    }
+}
+
+function applyFloatDomLayout(
+    wrapper: HTMLDivElement,
+    inner: HTMLDivElement,
+    layout: ReturnType<typeof resolveFloatDomLayout>
+): void {
+    const { wrapper: wrapperStyle, inner: innerStyle } = layout;
+    setStyleProperty(wrapper.style, 'top', `${wrapperStyle.top}px`);
```
jikkai: Write some code that looks like it was written by a human.

jikkai: Write some code that looks like it was written by a human.

packages/ui/src/index.ts
```diff
@@ -66,7 +66,7 @@ export type { IContextMenuHandler } from './services/contextmenu/contextmenu.ser
 export { DesktopDialogService } from './services/dialog/desktop-dialog.service';
 export { IDialogService } from './services/dialog/dialog.service';
 export { CanvasFloatDomPreviewService, CanvasFloatDomService } from './services/dom/canvas-dom-layer.service';
-export type { ICanvasFloatDomPreview, ICanvasFloatDomPreviewRequest, IFloatDom, IFloatDomLayout } from './services/dom/canvas-dom-layer.service';
+export type { ICanvasFloatDomPreview, ICanvasFloatDomPreviewRequest, IFloatDom, IFloatDomContentBoxConfig, IFloatDomLayout } from './services/dom/canvas-dom-layer.service';
```
jikkai: I don't think there is any place where this interface needs to be imported.

jikkai (changes requested):

packages/ui/src/views/components/dom/FloatDom.tsx
siam-ese: These values come from the runtime position$, so they can’t be represented by statically compiled Tailwind utilities.

jikkai (approved):

PR #7260, opened 2026-07-14
fix(ui): align title-only ribbon styling

DR-Univer:
## What changed

- Gives title-only ribbon overrides the same gray header background as the classic Ribbon menu.
- Uses the active Ribbon tab font weight, shadow, theme colors, and dark-mode colors.
- Extends the existing Ribbon override test to cover the shared styling contract.

## Why

Embedded tab/page blocks with title-only host chrome rendered their title on a white strip with medium-weight text, so switching to a Board block looked visually disconnected from the main Ribbon menu.

## Validation

- `@univerjs/ui`: 62 test files, 264 tests passed.
- Targeted ESLint: 0 errors.


DR-Univer:
Closing this change because Boards and Bases already share the same title-only Ribbon override. Their visual difference should be fixed in the Boards embed host path instead of changing the shared Ribbon placeholder styling.

PR #7273, opened 2026-07-16
fix(formula): preserve dirty ranges before initial calculation

DR-Univer:
## What changed

- instantiate the formula calculation trigger at `Ready` so it can collect dirty mutations during initialization
- defer browser-side flushing until `Rendered`, while starting immediately in Node.js
- merge server changeset dirty ranges and local cached mutation dirty ranges into the initial calculation
- keep the lifecycle decision in a controller instead of plugin business logic

## Why

A workbook is created before server changesets and locally cached mutations are replayed. The initial `WHEN_EMPTY` calculation can therefore happen before those mutations, while the formula trigger service may not yet be listening. Cached formula values then remain stale.

Local recovery mutations are replayed without `fromChangeset`, so the hold must preserve all dirty commands before the first calculation rather than filtering by that option.

## Impact

Dirty mutations produced before browser rendering are calculated once the formula engine starts. Normal post-render calculation behavior is unchanged, and Node.js starts without waiting for a render lifecycle.

## Validation

- `pnpm vitest run packages/engine-formula/src/services/__tests__/formula-calculation-trigger.service.spec.ts packages/engine-formula/src/controllers/__tests__/formula-calculation-trigger.controller.spec.ts packages/sheets-formula/src/controllers/__tests__/trigger-calculation.controller.spec.ts`
- `pnpm --filter @univerjs/sheets-formula typecheck`

### DIFF
--- a/packages/engine-formula/src/controllers/__tests__/formula-calculation-trigger.controller.spec.ts
+++ b/packages/engine-formula/src/controllers/__tests__/formula-calculation-trigger.controller.spec.ts
@@ -1,0 +1,58 @@
+/**
+ * Copyright 2023-present DreamNum Co., Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


packages/engine-formula/src/controllers/__tests__/formula-calculation-trigger.controller.spec.ts
jikkai, line 17: Barrel import is prohibited unless necessary

jikkai (changes requested):

PR #7287, opened 2026-07-17
fix(docs): respect layout fragment geometry

DR-Univer:
## What changed

- keep paragraph backgrounds and borders scoped to their current column
- restore section render offsets before advancing across pages
- keep positioned drawing anchors separate from flowing text
- refresh non-inline drawings after transform mutations and correct page-relative placement behavior
- let custom ranges provide ascent/descent metrics so formula glyphs participate in line-height calculation

## Why

Several document layout paths used page-wide or placeholder metrics after content entered a column or custom range. This caused decorations to span columns, drawing positions to drift, and tall formulas to receive insufficient line height.

## Impact

Traditional sections, nested column layouts, drawings, and formula placeholders now use their actual layout fragment geometry without special-casing table or column-group rendering.

## Validation

- `@univerjs/engine-render`: 81 targeted tests passed
- `@univerjs/docs-drawing-ui`: 36 targeted tests passed
- typecheck passed for `@univerjs/core`, `@univerjs/engine-render`, and `@univerjs/docs-drawing-ui`


packages/docs-ui/src/menu/__tests__/menu.spec.ts
```diff
@@ -0,0 +1,90 @@
+/**
```
jikkai: We don't need to test the menu.ts

jikkai (dismissed):

jikkai (dismissed):

PR #7306, opened 2026-07-22
fix(sheets-formula): update formula refs when deleting rows on inactive sheet

### LINKED ISSUE #7305: [Bug] Formula refs not updated when deleting rows on a non-active sheet

0.25.1

### Reproduction link

https://github.com/QJerryxjh/test_univer/tree/formual_move

When deleteRows runs on a sheet that is not active, cells move but formula references are not adjusted.

Repro

Sheet1 A4:C4 = [1, 2, '=sum(A4:B4)']
Switch to another sheet
sheet1.deleteRows(0, 2)
Expected: C2 = =sum(A2:B2)
Actual: C2 = =sum(A4:B4) (cell moved, formula unchanged)

If Sheet1 stays active during deleteRows, the formula updates correctly

### Expected behavior

the formula updates correctly when delete row or column not depend actvie sheet

### Actual behavior

When deleteRows runs on a sheet that is not active, cells move but formula references are not adjusted.

### System information

  System:
    OS: macOS 26.5.2
    CPU: (8) arm64 Apple M1
    Memory: 105.81 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.10.0 - /Users/qjerry/.nvm/versions/node/v20.10.0/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 10.2.3 - /Users/qjerry/.nvm/versions/node/v20.10.0/bin/npm
    pnpm: 10.5.2 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 150.0.7871.129
    Firefox: 151.0.4
    Safari: 26.5.2

QJerryxjh:
Closes #7305

## Summary

`FWorksheet.deleteRows()` / `deleteColumns()` (and the underlying `RemoveRowByRange` / `RemoveColByRange` commands) failed to update formula references when the mutated sheet was **not** the active sheet. The cell contents still shifted correctly (e.g. `C4` moved to `C2`), but formula strings kept their old offsets (e.g. stayed `=SUM(A4:B4)` instead of becoming `=SUM(A2:B2)`).

The remove-row/col interceptor path only forwarded `{ range }` into `SheetInterceptorService`, so formula update logic resolved the target via the **active** sheet (`getCurrentSheetInfo` / `getSheetCommandTarget` without params). When Sheet1 was mutated while Sheet2 was active, offsets were computed against the wrong sheet and Sheet1 formulas were left unchanged.

## Why

Facade APIs already pass explicit `unitId` / `subUnitId` into `RemoveRowByRangeCommand` / `RemoveColByRangeCommand`, and `handleRefInsertRow` / `handleRefInsertCol` already prefer those ids. The remove path disagreed: interceptors dropped the sheet identity, and `handleRefRemoveRow` / `handleRefRemoveCol` always fell back to the active sheet. That mismatch is the reported bug. The fix threads `unitId` / `subUnitId` through remove interceptors and ref-range consumers, and prefers explicit params with a fallback to the active sheet so the active-sheet UI path stays unchanged.

## How to test

1. Regression tests for the facade / non-active path:

```
pnpm --filter @univerjs/sheets-formula-ui exec vitest run \
  src/controllers/__tests__/update-formula.controller.spec.ts \
  -t "non-active sheet"
```

Covers:
- `Remove rows on non-active sheet, update reference and position`
- `Remove columns on non-active sheet, update reference and position`

2. Param mapping unit tests:

```
pnpm --filter @univerjs/sheets-formula exec vitest run \
  src/controllers/utils/__tests__/ref-range-param.spec.ts
```

3. Manual (optional): in `examples` (`pnpm dev` → `preset-sheets-core`), seed `Sheet1!A4:C4` with `1, 2, =SUM(A4:B4)`, switch to another sheet, call `sheet1.deleteRows(0, 2)`, then assert `Sheet1!C2` is `=SUM(A2:B2)`.

## Pull Request Checklist

### DIFF
--- a/packages/sheets-formula-ui/src/controllers/__tests__/update-formula.controller.spec.ts
+++ b/packages/sheets-formula-ui/src/controllers/__tests__/update-formula.controller.spec.ts
@@ -1432,6 +1432,44 @@ describe('Test update formula ', () => {
             expect(valuesRedo2).toStrictEqual([[{ f: '=SUM(A1:B1)' }], [{ v: 1, t: CellValueType.NUMBER }]]);
         });
 
+        // Facade deleteRows path: RemoveRowByRange with explicit unitId/subUnitId while another sheet is active
+        it('Remove rows on non-active sheet, update reference and position', async () => {
+            const workbook = get(IUniverInstanceService).getUnit<Workbook>('test');
+            const sheet2 = workbook?.getSheetBySheetId('sheet2');
+            if (!sheet2) {
+                throw new Error('sheet2 not found');
+            }
+            workbook?.setActiveSheet(sheet2);
+
+            expect(commandService.syncExecuteCommand(RemoveRowByRangeCommand.id, {
+                unitId: 'test',
+                subUnitId: 'sheet1',
+                range: {
+                    startRow: 1,
+                    endRow: 1,
+                    startColumn: 0,
+                    endColumn: 19,
+                },
+            })).toBeTruthy();
+
+            const values = getValues(1, 2, 2, 2, 'sheet1');
+            expect(values).toStrictEqual([[{ f: '=A1:B1' }], [null]]);
+            const values2 = getValues(4, 2, 5, 2, 'sheet1');
+            expect(values2).toStrictEqual([[{ f: '=SUM(A1:B1)' }], [{ v: 1, t: CellValueType.NUMBER }]]);
+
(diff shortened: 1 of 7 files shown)


jikkai:
Thanks for your PR. I have updated the PR so row and column deletions always use the target sheet instead of falling back to the active one, and cleaned up the related tests as well.

jikkai (approved):

PR #7311, opened 2026-07-23
feat(core): define Base attachment value type

DR-Univer:
## Summary
- add the shared IBaseAttachment domain type to Base core
- reuse the existing ImageSourceType enum for URL, UUID, and Base64 sources
- export the type from @univerjs/core for headless/model APIs

## Validation
- pnpm eslint packages/core/src/bases/index.ts packages/core/src/bases/typedef.ts
- pnpm --filter @univerjs/core typecheck

PR #7317, opened 2026-07-23
feat(core): define Base attachment value type

DR-Univer:
## What changed

- Define and export the shared Base attachment value model in `@univerjs/core`.
- Cover both embedded Base64 attachments and remotely uploaded attachment IDs with one canonical contract.

## Why

The Base model and facade need a headless-safe attachment value shape that does not depend on UI packages.

## Validation

- `pnpm --filter @univerjs/core typecheck`
- Final diff contains only the two core Base type/export files.
- No process fixtures, screenshots, images, examples, or documentation are included.

### DIFF
--- a/packages/core/src/bases/typedef.ts
+++ b/packages/core/src/bases/typedef.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import type { ImageSourceType } from '../services/image-io/image-io.service';
 import type { CellValueType, LocaleType } from '../types/enum';
 import type { IDocumentData } from '../types/interfaces';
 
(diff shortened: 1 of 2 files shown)


jikkai (changes requested):
Duplicate of #7311

packages/core/src/bases/typedef.ts
```diff
@@ -51,6 +52,29 @@ export interface IBaseResources {
...
+    /** Image width in pixels when known. */
+    width?: number;
+    /** Image height in pixels when known. */
+    height?: number;
+    /** How to interpret `source`, for example UUID, URL, or Base64 data. */
+    sourceType?: ImageSourceType;
+    /** Canonical attachment source: an uploaded file id, URL, or Base64 data URL. */
+    source?: string;
+    /** @deprecated Store Base64 data in `source` with `sourceType` set to `BASE64`. */
+    base64?: string;
```
jikkai: Why are there create-as-deprecated fields?

jikkai (changes requested):

jikkai (approved):

hexf00 (approved):

PR #7331, opened 2026-07-25
fix(engine-render): clip cache blit source rect to canvas bounds for Safari < 18

ssssssssssss:
## Summary

Frozen rows/columns lose their content on Safari < 18 when the viewport cache is enabled (reproduced with conditional-formatting data bars in a frozen column, but it affects all cell rendering in frozen viewports).

**Root cause.** `Spreadsheet.renderByViewports()` blits the frozen viewport's cache canvas back to the main canvas with a source rectangle larger than the cache canvas itself:

```ts
const dw = right - left + rowHeaderWidthAndMarginLeft;
const dh = bottom - top + columnHeaderHeightAndMarginTop;
```

Frozen viewports are created with `bufferEdgeX: 0` (viewMainLeft) / `bufferEdgeY: 0` (viewMainTop) in sheets-ui, so the cache canvas is exactly the viewport size on that axis and the blit source rect overflows it by the header margin (~46px horizontally for viewMainLeft, ~26px vertically for viewMainTop). viewMain is unaffected because its buffer edges (100px) exceed the margins — which is why only frozen areas break.

Per the [HTML spec](https://html.spec.whatwg.org/multipage/canvas.html#drawing-images), when the source rectangle is outside the source image it "must be clipped to the source image and the destination rectangle must be clipped in the same proportion", with the out-of-bounds area treated as transparent. Blink and Gecko implement this, so the overflow is invisible there. WebKit only adopted this behavior in Safari 18.0 ("Fixed drawImage to not alter the input source or the destination rectangles" — [WebKit Features in Safari 18.0](https://webkit.org/blog/15865/webkit-features-in-safari-18-0/)). On Safari < 18 the overflowed `drawImage` is dropped entirely or the clipped source is stretched into the full destination rect, so frozen viewports render empty or displaced.

**Fix.** Clip the source rectangle to the cache canvas bounds in `Spreadsheet._applyCache()` and shrink the destination rectangle proportionally — i.e. apply the spec behavior at the call site instead of relying on the browser. On spec-compliant browsers this produces the exact same pixels as before (the clipped-away region was transparent padding), and on Safari < 18 it restores frozen viewport rendering. The viewport cache stays enabled on all browsers.

## How to test

1. Create a spreadsheet, freeze the first column (and/or first row), and add content in the frozen area (e.g. values with conditional-formatting data bars).
2. Open in Safari 15–17: before this change the frozen area content is missing or misplaced; after this change it renders correctly. Scrolling performance is unchanged since the cache stays enabled.
3. Chrome/Firefox/Safari 18+: no visual change.

Unit tests in `packages/engine-render/src/components/sheets/__tests__/spreadsheet.apply-cache.spec.ts` cover right/bottom overflow clipping, proportional destination shrinking, negative source offsets, fully-out-of-bounds early return, and zero-size passthrough. Full `@univerjs/engine-render` suite passes (819 tests) and `tsc --noEmit` is clean.

## Pull Request Checklist


jikkai (approved):
LGTM

PR #7362, opened 2026-07-28
feat(ui): support bounded popups without animation

DR-Univer:
## Summary

- add opt-in `boundaryElement` and `disableAnimation` support to `RectPopup`
- constrain popup placement to a canvas boundary when requested
- keep an animation-disabled popup offscreen while async content still has a zero-size shell

## Why

The Slide shape toolbar could visibly move from right to left because its first render was positioned with zero width and then recentered after async content mounted. Consumers also need an opt-in way to keep aggregate-selection toolbars inside their canvas.

Default popup behavior remains unchanged unless these properties are supplied.

## Validation

- ESLint: 0 errors for the changed UI files
- `@univerjs/ui` typecheck passed
- `CanvasPopup.spec.tsx`: 3 tests passed

### DIFF
--- a/packages/ui/src/views/components/popup/RectPopup.tsx
+++ b/packages/ui/src/views/components/popup/RectPopup.tsx
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import type { Nullable } from '@univerjs/core';
+import type { ReactNode, RefObject } from 'react';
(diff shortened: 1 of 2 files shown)


DR-Univer:
Closing as requested: the RectPopup base-component changes and their test have been fully reverted. The animation issue will be handled separately.

PR #7373, opened 2026-07-30
fix(ui): retry emoji source downloads

DR-Univer:
## What changed

- Retry emoji source downloads up to three times for network failures, HTTP 408/429, and 5xx responses.
- Use exponential backoff and a 30-second timeout per attempt.

## Why

The OSS dev CI run https://github.com/dream-num/univer/actions/runs/30463319668 failed during `pnpm install --frozen-lockfile` because the emoji generator hit a transient jsDelivr `ETIMEDOUT`. A single CDN connection failure should not abort every test shard.

The merged Base change in #7372 is unrelated; typecheck and ESLint passed in the failed run.

## Validation

- `pnpm install --frozen-lockfile` including the real emoji generator
- `node --experimental-strip-types packages/ui/scripts/generate-emojis.mts`
- `pnpm --filter @univerjs/ui test` — 69 files, 326 tests passed
- `pnpm --filter @univerjs/ui typecheck`
- ESLint on the changed script
- `git diff --check`


packages/ui/scripts/__tests__/generate-emojis.spec.ts
```diff
@@ -0,0 +1,73 @@
+/**
```
jikkai: I don't think test cases are necessary here.

jikkai (changes requested):

jikkai (approved):

PR #7416, opened 2026-08-04
feat: improve undo redo history handling

DR-Univer:
## Summary

- raise the default per-unit Undo/Redo history limit from 20 to 50 command groups and expose `undoRedoHistoryLimit` through the core `Univer` configuration
- add a generic scoped grouping primitive to `IUndoRedoService`; group metadata stays inside the service and is not exposed to commands, mutations, or history items
- add the UI-owned `UndoRedoGroupService` for explicit multi-command actions and short direction-key nudge sessions
- coalesce repeated Sheets drawing nudges from the UI while keeping `SetSheetDrawingCommand` unaware of grouping
- keep ordinary property-panel edits independent; the grouping service does not infer property scopes

## Verification

- `@univerjs/core`: 148 test files, 986 tests passed
- `@univerjs/ui`: 69 test files, 332 tests passed
- final targeted Core grouping tests: 8/8 passed
- final targeted UI grouping tests: 2/2 passed
- targeted Sheets drawing tests: 7/7 passed
- Core and UI typecheck passed
- changed-file ESLint passed with no errors; `git diff --check` passed
- no demo implementation, docs, images, process-only tests, ESLint config, or build config changes

## Notes

- history stacks remain isolated per unit ID
- setting `undoRedoHistoryLimit` to `0` disables local Undo history
- UI code chooses explicit action boundaries; model commands continue to own normal mutation-based Undo/Redo only


packages/ui/package.json
```diff
@@ -27,6 +27,7 @@
     "exports": {
         ".": "./src/index.ts",
         "./*": "./src/*",
+        "./services/undo-redo/undo-redo-group.service": "./src/services/undo-redo/undo-redo-group.service.ts",
```
jikkai: WTF?

jikkai (changes requested):

packages/ui/package.json
DR-Univer: Replaced the internal-looking deep export with the first-class public subpath `@univerjs/ui/undo-redo-group`. A package entry is still required because the existing `./* -> ./src/*` mapping does not append `.ts`; removing the entry produces TS2307 in Pro consumers. No barrel re-export was added.

jikkai: No, new subpaths are never allowed.

PR #7444, opened 2026-08-08
fix(engine-formula): preserve declared LINEST array spills

DR-Univer:
## Summary

- preserve top-level legacy-array results when a declared array range is active
- keep normal single-cell LINEST formulas on the existing scalar compatibility path
- cover function-reference override lifecycle behavior

## Root cause

`FunctionNode` always reduced top-level legacy-array results to their first cell. The calculation runtime already supplied an explicit row/column override for declared array formulas, but the normalization step did not distinguish that case from a normal single-cell formula.

## Validation

- `pnpm exec vitest run src/services/__tests__/runtime.service.spec.ts src/functions/statistical/linest/__tests__/index.spec.ts` (22 tests passed)
- targeted ESLint: 0 errors
- `git diff --check`

## Pull Request Checklist

- [x] No related issue.
- [x] Naming conventions are followed.
- [x] Unit tests cover the change.

### DIFF
--- a/packages/engine-formula/src/engine/ast-node/function-node.ts
+++ b/packages/engine-formula/src/engine/ast-node/function-node.ts
@@ -206,6 +206,13 @@ export class FunctionNode extends BaseAstNode {
             return result;
         }
 
+        if (
+            this._runtimeService.hasFunctionRefInfoOverride() &&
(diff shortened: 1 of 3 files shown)


DR-Univer:
Folded into #7438 so the OSS formula changes remain in one PR.

PR #7446, opened 2026-08-08
fix(engine-formula): materialize union reference offsets

DR-Univer:
## Summary

- Materialize both cell references before combining them with the range operator.
- Prevent shared-formula offsets from being applied twice when one endpoint is already materialized by an address function such as `OFFSET`.
- Add regression coverage for mixed materialized references and ordinary matching shared offsets.

## Root cause

`CellReferenceObject.unionBy` combined raw range data and then copied the left reference offset to the resulting range. For formulas such as `=SUM($B$2:OFFSET(B2,0,0))`, `OFFSET` had already materialized its endpoint, so the shared-formula follower offset was applied a second time by the combined range.

## Impact

Shared-formula followers that combine a direct cell reference with an address-function result now read the same range as Excel. Ordinary relative range unions continue to apply their shared offset exactly once.

## Validation

- `pnpm --filter @univerjs/engine-formula exec vitest run src/engine/ast-node/__tests__/union-node.spec.ts src/functions/lookup/offset/__tests__/index.spec.ts src/engine/reference-object/__tests__/cell-reference-object.spec.ts` (25 tests passed)
- Dual-engine workbook snapshot suite in univer-rs (19 snapshots passed against Excel COM-verified caches)
- `git diff --check`

## Pull Request Checklist

- [x] Missing issue; this is a regression found by workbook-level compatibility snapshots.
- [x] Naming convention is followed.
- [x] Unit tests have been added.

### DIFF
--- a/packages/engine-formula/src/engine/reference-object/__tests__/cell-reference-object.spec.ts
+++ b/packages/engine-formula/src/engine/reference-object/__tests__/cell-reference-object.spec.ts
@@ -17,6 +17,7 @@
 import { CellValueType, ObjectMatrix } from '@univerjs/core';
 import { describe, expect, it } from 'vitest';
 import { CellReferenceObject } from '../cell-reference-object';
+import { RangeReferenceObject } from '../range-reference-object';
 
 describe('CellReferenceObject', () => {
     it('should allow blank cells inside the Excel grid even when they are outside snapshot row data', () => {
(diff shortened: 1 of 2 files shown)


DR-Univer:
Superseded by #7438 so the formula compatibility changes remain in one coordinated OSS PR.

PR #7480, opened 2026-08-11
test(sheets): stabilize merged-cell visual comparison

DR-Univer:
## Summary

- restore the merged-cell Linux visual baseline omitted by the runtime-scope revert in #7472
- replace the fixed post-load delay with explicit loading, font, and render-frame readiness checks

## Root cause

#7460 updated both runtime-scope behavior and the merged-cell snapshot. #7472 reverted the runtime behavior but left the newer snapshot in place. Current CI output is deterministic and byte-for-byte matches the baseline from before #7460.

## Validation

- restored snapshot SHA-256 matches both the failed CI artifact and the pre-#7460 baseline
- `pnpm exec eslint e2e/visual-comparison/sheets/sheets-visual-comparison.spec.ts`
- focused Playwright test completed the new readiness sequence and reached screenshot comparison locally; final Linux pixel comparison is delegated to CI because the repository has no macOS CI baseline
- `git diff --check`


DR-Univer:
Superseded by #7476, where the verified CI fix is now included.

PR #7492, opened 2026-08-12
fix(engine-formula): resolve cross-unit table references

hexf00:
## Why

Cross-Unit Base table references failed in two consecutive ownership boundaries:

1. Product Unit display names may contain `|`, while the table-reference matcher reused the stricter Excel workbook-name character set. `[Urban Nomad | Showcase]!Table[Column]` was therefore not recognized and evaluated to `#NAME?`.
2. `INDEX` materializes a table-column item as a cell reference with a direct Base table ID. `FunctionNode` then calls `setForcedSheetId()`; the method resolved an empty sheet name and cleared that direct table ID. Non-Base hosts subsequently read the current Doc/Slide subUnit and returned an empty value.

Related to dream-num/univer-pro#5495.

## Correctness

- Parse bracketed product display Unit qualifiers with Unit-name semantics while preserving OOXML `[n]!Table[Column]`, quoted, unquoted, and legacy runtime-ID forms.
- Resolve a forced sheet name only when one exists, preserving a directly assigned sheet/table ID.
- Cover both boundaries with focused matcher and reference-object regressions.

## Long-term correctness

The parser owns textual Unit qualifier recognition. `BaseReferenceObject` owns the distinction between a sheet name that still needs resolution and a direct sheet/table ID that is already resolved. The fix applies once in these shared boundaries for Sheet formulas and non-Sheet other formulas.

## Validation

- `vitest run packages/engine-formula/src/basics/__tests__/regex.spec.ts packages/engine-formula/src/engine/reference-object/__tests__/cell-reference-object.spec.ts`: 83 passed.
- `pnpm --filter @univerjs/engine-formula typecheck`: passed.
- Pro integration: 84 `CalculateFormulaProService` tests passed, including the original `INDEX` + `MATCH` Shape/other-formula structure; expected result `8`.

### DIFF
--- a/packages/engine-formula/src/basics/__tests__/regex.spec.ts
+++ b/packages/engine-formula/src/basics/__tests__/regex.spec.ts
@@ -80,6 +80,7 @@ describe('Test ref regex', () => {
         expect(regexTestSingeRange('[1]Sheet1!A1')).toBe(true);
         expect(new RegExp(REFERENCE_TABLE_SINGLE_COLUMN_REGEX).test('[Book]Sheet1!A1')).toBe(false);
         expect(new RegExp(REFERENCE_TABLE_SINGLE_COLUMN_REGEX).test('[1]!SalesTable[Amount]')).toBe(true);
+        expect(new RegExp(REFERENCE_TABLE_SINGLE_COLUMN_REGEX).test('[Urban Nomad | Showcase]!Inventory[Safety stock]')).toBe(true);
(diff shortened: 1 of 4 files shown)


hexf00:
Closing after revalidation. The failing case uses `[Unit | Name]!Table[Column]`, but `|` is not valid in the Excel-compatible external workbook qualifier represented by this syntax. With a valid qualifier, the original implementation passes the INDEX/MATCH regression without either change in this PR. The fix belongs in formula generation/validation or data migration, not in broadening the formula parser.

PR #7505, opened 2026-08-13
perf(sheets): accelerate canvas scrolling

DR-Univer:
## What changed

- Coalesce wheel input into one relative-scroll command per animation frame.
- Add a clean-scene scroll fast path that shifts the existing engine canvas and repaints only exposed strips and scrollbars.
- Alternate viewport cache canvases during incremental sheet rendering to avoid self-copy temporary buffers.
- Keep independent cache pairs per viewport so frozen rows and columns do not resize a shared buffer on every frame.
- Reuse the conditional-formatting rule range index to skip DataBar and Icon renderers when exposed scroll strips do not intersect matching rules.
- Intersect dirty ranges with matching DataBar/Icon rule ranges instead of scanning unrelated cells.
- Cache composed conditional-formatting results and invalidate them whenever rules, calculations, or the underlying view-model cache change.
- Reuse intercepted cell data from the sheet style cache when rendering markers instead of executing the cell interceptor chain again.
- Preserve the original conditional-formatting render path when the resolver is unavailable or merged cells make range clipping unsafe.
- Fall back to the existing full render path for dirty scenes, merged sheets, embedded scenes, and large scroll jumps.

## Why

Empty-sheet scrolling was still spending most of each frame copying and compositing full DPR-sized canvases. Dense sheets also invoked the DataBar and Icon extensions for every exposed cell even when no matching rule existed in the viewport. After those costs were removed, profiling showed that repeated conditional-formatting composition and repeated marker cell interception were the next avoidable costs. These changes reuse existing render metadata and keep invalidation at the models that own it.

## Impact

- The original clean-sheet profiling reduced hot scroll-frame rendering from roughly 23–30 ms to 19.5–25.4 ms.
- On the 20,000 × 48 mega-sales fixture at 1280 × 720 DPR 1, the first optimization pass increased scroll FPS from 45.6 to 71.3 and reduced mean dirty render time from 37.9 ms to 21.3 ms.
- In a same-commit A/B for the latest metadata reuse, scroll FPS increased from 70.6 to 75.1 and mean dirty render time decreased from 22.28 ms to 20.28 ms.
- Conditional-formatting composition decreased from 2.52 ms/frame to 1.51 ms/frame.
- Marker rendering decreased from 1.61 ms/frame to 0.07 ms/frame.
- The absent Icon extension cost decreased from 8.44 ms/frame to approximately 0 ms/frame.
- The off-viewport DataBar extension cost decreased from 8.66 ms/frame to approximately 0.01 ms/frame.
- Frozen and multi-viewport rendering remains covered with independent scroll regions and cache pairs.

## Validation

- `pnpm --filter @univerjs/engine-render test` (95 files, 952 passed, 59 skipped)
- `pnpm --filter @univerjs/sheets-conditional-formatting test` (15 files, 118 passed)
- `pnpm --filter @univerjs/engine-render typecheck`
- `pnpm --filter @univerjs/sheets-conditional-formatting typecheck`
- Targeted ESLint on the latest changed source and tests (0 errors; existing warnings only)
- `node scripts/check-conventions.mjs` from the Pro repository (completed with existing warnings)
- Browser verification after horizontal scrolling that DataBars, conditionally formatted cells, note markers, and the note popup remain visible with no page errors
- `git diff --check`

### Latest cache guard-band optimization

- Keep the existing 100 px safety overlap, but extend it only toward retained cache content instead of all four sides.
- Same-build A/B on the 20,000 × 48 fixture: mean dirty render time 20.49 ms → 15.64 ms (-23.7%); DefaultFontExtension 10.64 ms/frame → 6.54 ms/frame (-38.6%).
- Verified vertical, horizontal, diagonal, frozen-first-row, DPR 2, and 125% zoom scrolling.
- Baseline and optimized main canvases were pixel-identical after raw RGBA decoding at the same final scroll position.
- packages/engine-render: 95 test files passed, 952 tests passed, 59 skipped; typecheck passed.

### DIFF
--- a/packages/engine-render/src/__tests__/engine-scene-viewport.spec.ts
+++ b/packages/engine-render/src/__tests__/engine-scene-viewport.spec.ts
@@ -308,6 +308,67 @@ describe('engine scene viewport extra', () => {
         engine.dispose();
     });
 
+    it('preserves engine and layer caches only for a pending scroll render', () => {
+        const { engine, scene } = createFixture();
+        const layer = scene.getLayer(1);
+        const renderSpy = vi.spyOn(layer, 'render');
+        scene.render();
+        renderSpy.mockClear();
+
+        scene.makeDirtyForScrolling();
+        expect(scene.isScrollRenderPending()).toBe(true);
+        scene.render();
+        expect(scene.isScrollRenderPending()).toBe(false);
+        expect(renderSpy).toHaveBeenLastCalledWith(undefined, false, expect.objectContaining({
+            preserveCache: true,
+        }));
+
+        scene.makeDirtyForScrolling();
+        scene.makeDirty(true);
+        scene.render();
+        expect(renderSpy).toHaveBeenLastCalledWith(undefined, false, undefined);
+
+        scene.dispose();
+        engine.dispose();
+    });
+
+    it('scrolls independent viewport regions for a frozen sheet render', () => {
+        const { engine, scene, viewport, container } = createFixture();
+        viewport.setViewportSize({ left: 100, width: 200 });
+        const frozenViewport = new Viewport('viewMainLeft', scene, {
+            left: 0,
+            top: 0,
+            width: 100,
+            height: 180,
+            active: true,
+            allowCache: true,
+        });
+
+        scene.render();
+        const engineCtx = engine.getCanvas().getContext();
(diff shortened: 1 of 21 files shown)


DR-Univer:
Superseded by #7507, which consolidates all large-sheet scrolling performance changes into a single PR.

PR #7506, opened 2026-08-13
perf(sheets): accelerate canvas scrolling

DR-Univer:
## What changed

- Coalesce wheel input into one relative-scroll command per animation frame.
- Shift clean engine canvases and repaint only newly exposed scroll strips.
- Use alternating cache canvases per viewport, including frozen row and column viewports.
- Keep the existing 100 px safety overlap only toward retained cache content instead of expanding all four sides.
- Skip conditional-formatting renderers when dirty ranges do not intersect matching rules.
- Cache composed conditional-formatting styles with owner-level invalidation.
- Reuse intercepted cell data while rendering markers.
- Preserve the existing full-render fallback for dirty scenes, merged sheets, embedded scenes, and large jumps.

## Why

Scrolling dense sheets repeatedly copied full canvases and repainted cells outside the newly exposed region. Conditional-formatting and marker extensions also repeated work that had already been resolved for the same cells. The changes retain the established rendering and invalidation boundaries while reducing the amount of work performed per scroll frame.

## Impact

- On the 20,000 × 48 mega-sales fixture at 1280 × 720 DPR 1, the original optimization increased measured scroll throughput from 45.6 to 71.3 FPS and reduced mean dirty-render time from 37.9 ms to 21.3 ms.
- The latest guard-band A/B reduced mean dirty-render time from 20.49 ms to 15.64 ms and DefaultFontExtension time from 10.64 ms/frame to 6.54 ms/frame.
- Baseline and optimized main canvases were pixel-identical after raw RGBA decoding at the same final scroll position.

## Validation

- `pnpm test` in `packages/engine-render`: 95 files passed, 952 tests passed, 59 skipped.
- `pnpm typecheck` in `packages/engine-render`.
- `pnpm --filter @univerjs/sheets-conditional-formatting test`: 15 files, 118 tests passed.
- `pnpm --filter @univerjs/sheets-conditional-formatting typecheck`.
- Targeted ESLint: zero errors; existing warnings only.
- `node scripts/check-conventions.mjs` from the Pro repository: existing warnings only.
- Browser matrix: vertical, horizontal, diagonal, frozen first row, DPR 2, and 125% zoom.
- `git diff --check`.

### DIFF
--- a/packages/engine-render/src/__tests__/engine-scene-viewport.spec.ts
+++ b/packages/engine-render/src/__tests__/engine-scene-viewport.spec.ts
@@ -308,6 +308,67 @@ describe('engine scene viewport extra', () => {
         engine.dispose();
     });
 
+    it('preserves engine and layer caches only for a pending scroll render', () => {
+        const { engine, scene } = createFixture();
+        const layer = scene.getLayer(1);
+        const renderSpy = vi.spyOn(layer, 'render');
+        scene.render();
+        renderSpy.mockClear();
+
+        scene.makeDirtyForScrolling();
+        expect(scene.isScrollRenderPending()).toBe(true);
+        scene.render();
+        expect(scene.isScrollRenderPending()).toBe(false);
+        expect(renderSpy).toHaveBeenLastCalledWith(undefined, false, expect.objectContaining({
+            preserveCache: true,
+        }));
+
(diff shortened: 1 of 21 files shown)


DR-Univer:
Superseded by #7507, which consolidates all large-sheet scrolling performance changes into a single PR.

PR #7526, opened 2026-08-16
feat: support slide thumbnail menu styles

DR-Univer:
## Summary

- Allow DropdownMenu radio options to forward a custom className to the underlying radio item.
- Keep the API backward-compatible for existing radio options.

## Validation

- pnpm --filter @univerjs/design exec vitest run src/components/dropdown-menu/__tests__/index.spec.tsx (1 file, 8 tests passed)
- pnpm --filter @univerjs/design typecheck
- Existing commit hooks validated the changed TypeScript files

## Scope Notes

- The nested-workspace demo CSS workaround and its regression test are intentionally excluded from this PR and retained only for local demo verification.

## Pull Request Checklist

- [x] No related Issue was provided.
- [x] Naming conventions are followed.
- [x] Focused regression tests cover the public API change.


common/shared/esbuild/index.ts
```diff
@@ -25,12 +25,17 @@ interface IEsbuildPlugin {
     setup: (build: IPluginBuild) => void;
 }
 
+function isWorkspacePackageSource(importer: string): boolean {
+    const normalizedImporter = importer.replaceAll('\\', '/');
```
jikkai: Why modify it?

common/shared/esbuild/index.spec.ts
```diff
@@ -0,0 +1,34 @@
+import { describe, expect, it } from 'vitest';
```
jikkai: Why add tests here?

jikkai (changes requested):

common/shared/esbuild/index.ts
DR-Univer: The plugin receives an absolute importer path. The previous importer.includes("packages") check also matched an ancestor directory named packages, so examples/src/global.css was incorrectly treated as package CSS and replaced with an empty module in this nested workspace. Checking the normalized path segment /packages/ keeps the intended package-source behavior while allowing example-level global CSS to load.

common/shared/esbuild/index.spec.ts
DR-Univer: This test is regression coverage for the importer classification above. It verifies actual packages/* sources remain ignored, examples/* sources are not ignored even when an ancestor directory is named packages, Windows separators are normalized, and entry-point CSS without an importer remains unaffected. The broad substring check would fail the nested-workspace case.

common/shared/esbuild/index.ts
DR-Univer: Agreed on PR scope. I removed this esbuild workaround from the submitted diff and kept it only as an uncommitted local change for the nested-workspace demo.

common/shared/esbuild/index.spec.ts
DR-Univer: Agreed. The esbuild regression test is no longer part of this PR and remains local-only with the demo workaround.

jikkai:
The styling under the dropdown is unified and should not be overridden by any external styles. Try other solutions to solve the problem.

PR #7537, opened 2026-08-17
fix(engine-render): resolve theme colors when printing

DR-Univer:
## Summary

- forward the canvas color service to printing contexts
- keep theme-token colors consistent between editing and printing
- add regression coverage for theme-color resolution in printing mode

## Context

Board print and image export use the dedicated printing canvas. Without the color service, theme color tokens were assigned directly to the native canvas context. Invalid color assignments retained the previous fill style, changing element backgrounds and hiding text in exported output.

## Validation

- `pnpm exec eslint packages/engine-render/src/canvas.ts packages/engine-render/src/__tests__/canvas.spec.ts`
- `pnpm --filter @univerjs/engine-render test -- canvas.spec.ts` (95 files, 962 tests)
- `pnpm --filter @univerjs/engine-render build:types`
- `git diff --check`

### DIFF
--- a/packages/engine-render/src/canvas.ts
+++ b/packages/engine-render/src/canvas.ts
@@ -91,7 +91,9 @@ export class Canvas {
         }
 
         if (props.mode === CanvasRenderMode.Printing) {
-            this._context = new UniverPrintingContext(context);
+            this._context = new UniverPrintingContext(context, {
+                canvasColorService: props.colorService,
+            });
         } else {
             this._context = new UniverRenderingContext(context, {
                 canvasColorService: props.colorService,
@@ -161,10 +163,7 @@ export class Canvas {
         if (this._width === 0 || this._height === 0) {
             return;
         }
-        if (pixelRatio < 1) {
-            pixelRatio = 1;
-        }
-        this.setSize(this._width, this._height, pixelRatio);
+        this.setSize(this._width, this._height, Math.max(1, pixelRatio));
     }
 
     dispose() {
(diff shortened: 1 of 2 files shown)


packages/engine-render/src/canvas.ts
```diff
@@ -191,11 +190,11 @@ export class Canvas {
             // If this call fails (due to browser bug, like in Firefox 3.6),
             // then revert to previous no-parameter image/png behavior
             return this.getCanvasEle().toDataURL(mimeType, quality);
-        } catch (e) {
+        } catch {
             try {
                 return this.getCanvasEle().toDataURL();
             } catch (err: unknown) {
-                const { message } = err as Error;
+                const message = err instanceof Error ? err.message : String(err);
```
jikkai: Over defensive code

packages/engine-render/src/canvas.ts
```diff
@@ -191,11 +190,11 @@ export class Canvas {
             // If this call fails (due to browser bug, like in Firefox 3.6),
             // then revert to previous no-parameter image/png behavior
             return this.getCanvasEle().toDataURL(mimeType, quality);
-        } catch (e) {
+        } catch {
             try {
                 return this.getCanvasEle().toDataURL();
             } catch (err: unknown) {
```
jikkai, line 199: There is a wrong error type here.

jikkai (changes requested):

packages/engine-render/src/canvas.ts
DR-Univer: Fixed in da9fd456da. I removed the unrelated error-handling cleanup entirely; the PR now changes only the printing-context color-service wiring and its regression test.

packages/engine-render/src/canvas.ts
DR-Univer: Fixed in da9fd456da. I restored the existing catch behavior so this PR no longer changes or annotates the error type. The runtime diff is now limited to passing the color service into UniverPrintingContext.

jikkai (approved):

